PR #1, opened 2021-04-22
siac: Add --insecure-input flag for wallet unlock

Eriner:
Sometimes you're alone and not worried about anyone peeking over your
shoulder. In such cases, masking and hiding the user input is of little
value. The only non-shoulder-surf security mechanism this would disable
is the Apple Secure Keyboard Entry. This commit does not change the
default behavior.

Other notes, I replaced the "golang.org/x/crypto/ssh/terminal"
dependency with its indirect import, "golang.org/x/term", as the
former is deprecated in favor of the latter.

This feature was requested on Reddit and seemed rather sane: https://www.reddit.com/r/siacoin/comments/mvmq93/feature_request_display_user_input_at_the_wallet/

There may also be a better name than "insecure-input". "no-shoulder-surf" didn't sound as descriptive...

### DIFF
--- a/cmd/siac/walletcmd.go
+++ b/cmd/siac/walletcmd.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -13,7 +14,7 @@ import (
 
 	"github.com/spf13/cobra"
 	mnemonics "gitlab.com/NebulousLabs/entropy-mnemonics"
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 
 	"gitlab.com/NebulousLabs/encoding"
 	"gitlab.com/NebulousLabs/errors"
@@ -209,9 +210,11 @@ use it instead of displaying the typical interactive prompt.`,
 
 const askPasswordText = "We need to encrypt the new data using the current wallet password, please provide: "
 
-const currentPasswordText = "Current Password: "
-const newPasswordText = "New Password: "
-const confirmPasswordText = "Confirm: "
+const (
+	currentPasswordText = "Current Password: "
+	newPasswordText     = "New Password: "
+	confirmPasswordText = "Confirm: "
+)
 
 // For an unconfirmed Transaction, the TransactionTimestamp field is set to the
 // maximum value of a uint64.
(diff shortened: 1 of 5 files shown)


cmd/siac/walletcmd.go
```diff
@@ -220,9 +223,31 @@ const unconfirmedTransactionTimestamp = ^uint64(0)
...
+			os.Stdin,
+			os.Stdout,
+		}
+		t := term.NewTerminal(screen, "")
+		pw, err := t.ReadLine()
+		if err != nil && err != io.EOF {
+			return pw, err
+		}
+		fmt.Println()
+		return string(pw), nil
```
lukechampine, line 245: Surely this works just as well?
```go
pw, err := bufio.NewReader(os.Stdin).ReadString('\n')
````

File: cmd/siac/walletcmd.go [hunk 2]
```diff
@@ -209,9 +210,11 @@ use it instead of displaying the typical interactive prompt.`,
...
 const askPasswordText = "We need to encrypt the new data using the current wallet password, please provide: "
 
-const currentPasswordText = "Current Password: "
-const newPasswordText = "New Password: "
-const confirmPasswordText = "Confirm: "
+const (
+	currentPasswordText = "Current Password: "
+	newPasswordText     = "New Password: "
+	confirmPasswordText = "Confirm: "
+)
```
lukechampine, line 217: In general PRs (and especially commits) shouldn't make unnecessary stylistic changes. If we want to tweak the style, we can apply it project-wide in a separate PR.

cmd/siac/walletcmd.go
Eriner: But does that handle the `\r` for windows users?

cmd/siac/walletcmd.go
Eriner: My lint-on-save picked this one up. I guess I can edit the file in `nano` and revert it without the auto-linter overriding me.

lukechampine: might as well leave it in this case, but yeah going forward I'd rather not have random changes added by your linter

cmd/siac/walletcmd.go
lukechampine: I think if we do `strings.TrimSpace` (which we should probably do anyway) it'll be fine

cmd/siac/walletcmd.go
Eriner: no promises that I won't miss some, but I'll check in the future.

Eriner:
Addressed comments and force-pushed.

lukechampine:
Thanks!

PR #2, opened 2021-04-24
Add dockerfile

n8maninger:
I figured we could get the ball rolling. Automated builds could either use GitHub's repo or DockerHub. I'm personally more in favor of DockerHub because of its first party nature and additional exposure, but it would require setting up an additional Sia Foundation org there and using GitHub secrets. 

To do:
- [x] Create Dockerfile
- [x] Automated build + push using buildx

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -1,0 +1,35 @@
+name: Docker Publish
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on new SemVer tags
+  push:
+    tags: 
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+'
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+      - uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: crazy-max/ghaction-docker-meta@ae431178c17085afd420fdf7ca88b37bd32e3d7d
+        name: generate tags
+        id: meta
+        with:
+          images: ghcr.io/siafoundation/siad
+          tags: |
+            type=semver,pattern={{version}}
+      - uses: docker/build-push-action@v2
+        with:
+          context: .
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+# build sia
+FROM golang:1.16-alpine AS build
+
+WORKDIR /app
+
+COPY . .
+
+# need to run git status first to fix GIT_DIRTY detection in makefile
+RUN apk update \
+	&& apk add --no-cache git make ca-certificates \
+	&& update-ca-certificates \
+	&& git status > /dev/null \
+	&& make static
+
+FROM alpine:latest
+
+COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=build /app/release /
+
+# gateway port
+EXPOSE 9981/tcp
+# host RHP2 port
+EXPOSE 9982/tcp
+# host RHP3 port
+EXPOSE 9983/tcp
+
+# SIA_WALLET_PASSWORD is used to automatically unlock the wallet
+ENV SIA_WALLET_PASSWORD=
+# SIA_API_PASSWORD sets the password used for API authentication
+ENV SIA_API_PASSWORD=
+
+VOLUME [ "/sia-data" ]
+
+ENTRYPOINT [ "/siad", "--disable-api-security", "-d", "/sia-data", "--api-addr", ":9980" ]
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,9 @@ dev:
 dev-race:
 	GORACE='$(racevars)' go install -race -tags='dev debug profile netgo' -ldflags='$(ldflags)' $(pkgs)
 
+static:
+	go build -o release/ -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs)
+
 # release builds and installs release binaries.
 release:
 	go install -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs)


lukechampine:
I assume we can start with just GitHub, and add DockerHub later? What's the impact on the end-user? 

For GitHub Actions, you can probably use this as a template: https://github.com/lukechampine/walrus/pull/20

Tagging other people who were interested in a Sia Dockerfile: @Eriner @acejam

Eriner:
Is this the start of the deprecation of https://github.com/NebulousLabs/docker-sia ?

If so, I would suggest we take the `Dockerfile` in this PR and rename it to `Dockerfile.dev`, with the intent of creating a slimmer production `Dockerfile` (base image on `SCRATH` or `alpine` if `SCRATCH` is too troublesome).

I haven't begun to package `siad` into a SCRATCH container yet, happy if you beat me to it @n8maninger :)

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+WORKDIR /app
+
+COPY . .
+
+ENV GOBIN=/app/bin
+
+# updates the cache to fix GIT_DIRTY detection in makefile
+RUN git status > /dev/null
+RUN make release
```
Eriner, line 12: This directive and the one above could be combined with `&&`

n8maninger:
> If so, I would suggest we take the `Dockerfile` in this PR and rename it to `Dockerfile.dev`, with the intent of creating a slimmer production `Dockerfile` (base image on `SCRATH` or `alpine` if `SCRATCH` is too troublesome).

I'm definitely for switching to alpine at least, `siacentral/sia` is based on alpine. Scratch or distroless is easy enough with Go binaries, but I would be concerned about breaking workflows for people coming from `siacentral/sia` or `nebulouslabs/sia` with minimal benefit.

Eriner:
>  I would be concerned about breaking workflows for people coming from `siacentral/sia` or `nebulouslabs/sia` with minimal benefit.

What do you anticipate breaking? I would think the switch from glibc -> musl would be (potentially) more disruptive than anything. For those who need shell, either use the `dev` image I suggested above or wrap it:

```Dockerfile
FROM alpine:orwhatever

COPY --from=ghcr.io/siafoundation/siad:latest /app/bin/siad

# do stuff
ENTRYPOINT [ "/app/bin/siad"]
```

n8maninger:
I doubt most users are even aware of glibc vs musl lol, but you're right not providing a shell shouldn't be too disruptive 🙂

Eriner:
My stance is that in production, you shouldn't have a shell available. Containers are purpose-built and highly specific. 99% of cases you're simply naming the container to be used, mounting a data & config volume/envvars, and you're done. At least, until there is an *application* problem, at which point a container with a shell is (sometimes) needed. 

I maintain that we should provide a `Dockerfile.dev` w/ the debian container from your original PR, or something close to it. I reviewed the existing PR and sent a PR to your branch @n8maninger.

acejam:
@Eriner Most users will need a shell when running `siad` in production. There are plenty of logs that are not sent to `stdout` (for good reason) and simple things like checking allowances and contracts require the `siac` CLI. `exec`'ing each command directly into the container is likely to be more painful. The `Alpine` / `ash` flavor that we've been using for the past few years has served us well.

Eriner:
> There are plenty of logs that are not sent to `stdout` (for good reason) 

This is what volume mounts are for.

> simple things like checking allowances and contracts require the `siac` CLI. 

Servers and clients don't have much business running in the same container. One container, one process. I have nothing against building and including the `siac` binary in this image, but that is not a convincing argument as to why a shell is necessary. Override the entrypoint, wrap the container, etc. Lots of options.

> `exec`'ing each command directly into the container is likely to be more painful.

Nobody should be `exec`ing anything in production. The `Dockerfile.dev` I've advocated for above should absolutely have a shell.

.github/workflows/publish-docker.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+      - uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: crazy-max/ghaction-docker-meta@v2
```
Eriner, line 23: I still don't think this should be in the image build pipeline and instead should be in a separate (dependent) step, only passing the output.

acejam:
> This is what volume mounts are for.

I'm well aware of what volume mounts are for, this container already defines one as you know. The purpose behind my comment is that a busy Sia node generates a decent amount of log files, all from separate internal modules. (gateway, contractor, renter, etc) Dumping them all to `stdout` would likely create an unpleasant stream to grok.

It's also going to be hard to read those mounted files on the host machine directly, since `siad` runs as root, and you'd never log into a production sever as root to read a root-owned file, right? 😄 (Thank you, sudo)

Speaking of which, this image runs as root, so we're already violating probably the most important Docker best practice that there is.

> Servers and clients don't have much business running in the same container. One container, one process.

There are two sides in the Docker camp - one container, one process - and - one container, _one logical service_, or _responsibility_. It is the latter approach that the Sia community has adopted and has been following for the last few years.

If we want this image to be widely used, any Dockerfile we add to this repo should ideally match https://github.com/NebulousLabs/docker-sia/ somewhat closely for compatibility reasons. I'm sure there are a few tweaks worth making though.

> Nobody should be execing anything in production. 

If you try to put these rules into place, and truly try to uphold them, then you're going to have a very bad time running Sia. As you point out, a shell is great for debugging application platforms. I have news for you - Sia still has many application problems. A shell is pretty much required. I'd like to see you spin up a new scratch-based image with no shell and set your renter allowance. Not everyone has the guts to build applications that consume the Sia API like we do. Flexibility is key.

Nearly every "official" Docker image supports `exec`'ing and shells, so it's clear what the views are from the Docker developers themselves.

Eriner:
Re: root, it is not the job of the container to define runtime security. That is to say, it is the job of the orchestrator to assign RunsAsUser and similar security contexts. `USER` is a poor Dockerfile directive, and never should have existed.

> Dumping them all to stdout would likely create an unpleasant stream to grok

I don't understand what the objection here is, is there a specific change that you recommend?

> There are two sides

and like with most things, there is a right side and a wrong side. (I kid, kind of) 🙃 

> Nearly every "official" Docker image supports exec'ing and shells, so it's clear what the views are from the Docker developers themselves.

Every container supports `exec`ing, its just a matter of what is in the container to be `exec`d. Nothing is stopping you from execing `siad`, `siac`, or any other binaries in the image. I'm just arguing that those two aforementioned things are the only things that need to be in the image. I have no problem with there being a `-alpine` tag, but I remain unconvinced that a shell is necessary for a production image.

> Flexibility is key.

See my post above about wrapping the container. If the smallest possible container is the source (the one proposed), wrapping it in your own container to add shells, debugging scripts, or anything else is quite simple. Again, this is not to say that a `Dockerfile.dev` shouldn't be provided - it should.

lukechampine:
If we start with a minimal container, as @Eriner suggests, could the Dockerfiles in `docker-sia` be modified to wrap that container? I guess I'm just a little wary of adding everything in `docker-sia` to this repo. I mean, it has *seven* distinct Dockerfiles. Is that necessary?

Eriner:
I've never used them, so my comment doesn't mean much, but it looks like it. The only caveat being we would want to build and distribute static binaries for other architectures (as in, provide `/armv8/sia{d,c}` in addition to the default amd64 binaries).

acejam:
> Re: root, it is not the job of the container to define runtime security. That is to say, it is the job of the orchestrator to assign RunsAsUser and similar security contexts. `USER` is a poor Dockerfile directive, and never should have existed.

Orchestrators can override that, sure - but the `USER` directive is important and has been around since day one. The same goes for permissions. It shouldn't be ignored just because you don't like it. Surely you must care about things being somewhat secure by default? Case in point:

```
$ docker run -u 100 siacentral/sia
[failed to get API password; mkdir /.sia: permission denied]
```

Can you add yet another `&&` or `chown` layer in there and dupe the size of the image? Sure, or you can use the `USER` directive and set permissions and the default `uid` at the same time, which is how it was intended to be used.

> If we start with a minimal container, as @Eriner suggests, could the Dockerfiles in `docker-sia` be modified to wrap that container? I guess I'm just a little wary of adding everything in `docker-sia` to this repo. I mean, it has _seven_ distinct Dockerfiles. Is that necessary?

@lukechampine It's not necessary. The existence of the `docker-sia` repo in itself is the problem. The Dockerifle should live with, be versioned with, and be released with the code it encapsulates. (containerizes) I'd copy over the `Alpine` variant Dockerfile, include whatever else is needed, and call it a day.

Eriner:
> Can you add yet another && or chown layer in there and dupe the size of the image? Sure, or you can use the USER directive and set permissions and the default uid at the same time, which is how it was intended to be used.

Or you could do neither of those and use the orchestrator to handle it at runtime. When running many containers in a cluster, there are likely only few you "own" from an application standpoint. For tasks like: managing volumes and data shared across containers, file permissions, or ensuring containers don't run as root or clobber host `uid`s, of the myriad of reasons to manage it an orchestrator is operator sanity. apparmor, uid, syscall access, etc. all should be controlled by the orchestrator at run-time, not at build-time. I maintain my stance that the `USER` directive is/was a mistake, as it has accustomed people to defining a security constraint at an incorrect "location" (time?).

Consider this, if a `Dockerfile` expects to operate with a specific UID/GID and `chowns` a bunch of files, "fixing" that image can often be annoying (I'm sure you've run into this at least once). Given the `siacentral` example you referenced above, often the solution is to mount a tmpfs if you don't care about the data, or a persistent volume if you do. Can be done at runtime, and much easier than the alternate scenario I outlined, fixing/altering permissions of files baked into the image. At least, that has been my experience.

.github/workflows/publish-docker.yml
n8maninger: To me it seems to unnecessarily complicate the pipeline with no benefit.

Eriner: If Mr Max decides one day in the future that he wants to modify the action to swap the `/bin/docker` binary with one that contains malicious code to generate altered images, that would be a problem. I'd suggest pinning the git hash, but that is even more annoying than splitting out the job. The issue can be avoided by moving it out of the (image) build path and into its own job, passing only the output. There is no shared machine/environment state in that case, whereas there is shared state currently.

n8maninger:
> I mean, it has seven distinct Dockerfiles. Is that necessary?

I don't think there's any need or desire to maintain more than 1 dockerfile here. A shell is really helpful for Sia, but not necessary. If we're going to include one, Alpine adds pretty minimal bloat.

acejam:
> > Can you add yet another && or chown layer in there and dupe the size of the image? Sure, or you can use the USER directive and set permissions and the default uid at the same time, which is how it was intended to be used.
> 
> Or you could do neither of those and use the orchestrator to handle it at runtime. When running many containers in a cluster, there are likely only few you "own" from an application standpoint. For tasks like: managing volumes and data shared across containers, file permissions, or ensuring containers don't run as root or clobber host `uid`s, of the myriad of reasons to manage it an orchestrator is operator sanity. apparmor, uid, syscall access, etc. all should be controlled by the orchestrator at run-time, not at build-time. I maintain my stance that the `USER` directive is/was a mistake, as it has accustomed people to defining a security constraint at an incorrect "location" (time?).
> 
> Consider this, if a `Dockerfile` expects to operate with a specific UID/GID and `chowns` a bunch of files, "fixing" that image can often be annoying (I'm sure you've run into this at least once). Given the `siacentral` example you referenced above, often the solution is to mount a tmpfs if you don't care about the data, or a persistent volume if you do. Can be done at runtime, and much easier than the alternate scenario I outlined, fixing/altering permissions of files baked into the image. At least, that has been my experience.

None of the Sia containers I've orchestrated have ever mounted the `~/.sia` path to begin with. The example above shows how simply changing the `uid` at runtime causes an error - nothing more. Mounting a `tmpfs` volume is yet-another-workaround, in my opinion.

With that said, all of my comments in this thread are marketing speak - we don't use this image anymore, and haven't for quite some time. We maintain our own private image. Advanced users are going to do the same.

Put your marketing hat on for a second: This image is likely to be the natural starting point for someone who wants to get started renting or hosting on Sia. It's no secret that Sia can be hard to use at times. **It has to be as easy to use as possible**. If you rip out the shell, people are going to throw their hands up, forget about Sia, and walk away. You've now lost a potential host, who didn't feel like reading the API docs, and simply wanted to use `siac`, or do something else with a shell. Most people who are using this image are going to simply `docker run` it, nothing more elaborate than that.

.github/workflows/publish-docker.yml
n8maninger: That's one of the reasons I was suggesting building custom tooling in the first place. If that's the concern, I would much rather pin it to a specific SHA. Adding a dependent job doesn't completely block malicious intent.

lukechampine:
> I don't think there's any need or desire to maintain more than 1 dockerfile here.

Can you explain briefly why we're able to get away with a single Dockerfile when `docker-sia` needs so many? If `docker-sia` disappeared, would it be a problem that `siad` only has one Dockerfile?

>I agree that a shell is really helpful for Sia. If we're going to include one, Alpine adds pretty minimal bloat and supports user's existing workflows.

Ok. This sounds like the most reasonable path forward. Personally, I feel the most aligned with the minimalist approach that @Eriner is advocating, but my opinion is moot because I don't use Docker myself. Given that @n8maninger and @acejam have been using Sia + Docker for years now, I trust their judgment the most. We can explore a slimmer image (either as an alternative or as the new default) later on.

n8maninger:
> Can you explain briefly why we're able to get away with a single Dockerfile when docker-sia needs so many? If docker-sia disappeared, would it be a problem that siad only has one Dockerfile?

The original `mtlynch/sia` image relied on downloading the binaries over building the code from source, but only supported `amd64`. The `nebulouslabs/sia` container has the same restraint while also supporting multiple architectures.

This image is based on `siacentral/sia`, which was built as a drop-in replacement for `mtlynch/sia` and consequently `nebulouslabs/sia`. Because it builds from source we can more easily support multiple architectures without separate tags or files.

If docker-sia disappeared, most users would only have to switch one line from `nebulouslabs/sia:latest` to `ghcr.io/siafoundation/siad:latest` and everything else would work the same.

I also agree that the minimalist container is technically better, but I think it's more important in this case to maintain the flexibility and compatibility for the broader community going forward. Alpine's surface area is minimal and the container size is only bloated to 33MiB over 27MiB.

lukechampine:
Also -- how does this all relate to CI? Will we use this Dockerfile as part of our automated testing, or will it just be used to build images after the test suite passes?

n8maninger:
> Also -- how does this all relate to CI? Will we use this Dockerfile as part of our automated testing, or will it just be used to build images after the test suite passes?

It's not directly related, the current workflow will only build and publish new docker images when a new release is cut.

I started to add testing workflows, but the makefile's `lint` target is currently failing which causes `test-long` to fail as well. Do you want to break that into its own PR or knock it all out in this one.

Eriner:
@n8maninger if it only a matter of timing out, you can increase the job timeout with `job-timeout` iirc.

n8maninger:
It's not just a matter of timing out. Even with the maximum of `timeout-minutes: 360` it's never successfully completed. Additionally, a few of the tests are race-y and randomly fail. Being able to re-run specific jobs within the workflow on GitLab was nice in those situations instead of having to rerun Sia's full hour+ testing suite.

I guess we could use the `test` target for new pr's, which completes in under a minute, and only use `test-long` for merges or on a nightly schedule?

Any thoughts? @lukechampine

lukechampine:
> Do you want to break that into its own PR or knock it all out in this one?

nah, I suppose it should be a separate PR.

> I've also never gotten test-long to successfully complete on GitHub's runners

The test suite has been expanding out of control for years now, and I'd really like to get a better handle on it. Perhaps we can start by ruthlessly disabling the worst offenders until the run time is more reasonable.

PR #7, opened 2021-05-18
Fix build.Critical triggering in uploadchunkdistributionqueue.go

ChrisSchinnerl:
This MR is meant to fix the Critical in `uploadchunkdistributionqueue.go` triggering in production.

The first commit adds a test to reproduce the Critical. It turned out that the panic may be triggered simply due to the structure of the locking in `threadedProcessQueue` and that there is no developer error.

That's why the second commit simply removes the sanity check.


lukechampine:
Thanks!

Issue #6, opened 2021-05-14
Expired data not being cleaned up


CtrlAltDefeat94:
I am running a script that runs through all contracts (API /host/contracts) and adds up all `datasize` for all the contracts with an `expirationheight` above the current block.

The result of this comes to a little below 7TB, whereas `siac host` states nearly 10TB of data is being used.
Semi recently, I've moved data around by resizing. 

Is my method of calculating my total used storage incorrectly, of could the resizing I've done be the cause? 
And if I'm correct that I'm still hosting data that should've been removed. How would I clear this up?

PR #9, opened 2021-05-24
Queue unresolved contracts past their proof deadline

n8maninger:
Changes `PruneStaleStorageObligation` to also queue action items for any unresolved contracts past their proof deadline when starting the host. 

Works around #6. The host from the issue has 487 unresolved contracts past their proof deadline. This will trigger data cleanup and mark those contracts as failed.


lukechampine:
nice!

n8maninger:
> nice! Is this sufficient to close #6, or is there more work to be done?

Yeah, this should be enough to close #6, but it won't fix the underlying issue with the contracts not resolving. I'm not sure how long that would take to find and fix. It might be a good idea to open a tracking issue for all the little Host bugs.

+ Contracts not being resolved
+ Lockups
+ Financial tracking

PR #10, opened 2021-05-25
Improve Registry Update

ChrisSchinnerl:
This MR contains various improvements to the registry update instruction.

1. It pulls a few pieces of code out into helper functions
2. Custom error for when an entry can't be updated due to insufficient work
3. ErrSameRevNum is only returned when the entries have the same revision and pow. This means they are considered equal. The worker won't consider the update a failure in that case.
4. fixes data race in hdb to get tests to pass

### DIFF
--- a/modules/host/contractmanager/sectorupdate_test.go
+++ b/modules/host/contractmanager/sectorupdate_test.go
@@ -2069,6 +2069,9 @@ func TestAddVirtualSectorOverflow(t *testing.T) {
 	if loadedOverflow != 1 {
 		t.Fatal("wrong overflow", loadedOverflow)
 	}
+	if err := loaded.Close(); err != nil {
+		t.Fatal(err)
+	}
 
 	// Remove the sector.
 	err = cmt.cm.RemoveSector(root)
@@ -2100,6 +2103,9 @@ func TestAddVirtualSectorOverflow(t *testing.T) {
 	loadedOverflow, exists = loaded.Overflow(id)
 	if !exists || loadedOverflow != 0 {
 		t.Fatal("overflow entry should be 0", loadedOverflow)
+	}
+	if err := loaded.Close(); err != nil {
+		t.Fatal(err)
 	}
 
 	// Create multiple threads, all adding sectors at the same time.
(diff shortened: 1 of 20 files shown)


modules/registry.go
```diff
@@ -79,6 +93,41 @@ func (entry RegistryValue) Sign(sk crypto.SecretKey) SignedRegistryValue {
...
+// ShouldUpdateWith returns true if entry2 would replace entry1 in a host's
+// registry. It returns false if it shouldn't. An error might be returned in
+// that case to specify the reason.
+func (entry *RegistryValue) ShouldUpdateWith(entry2 *RegistryValue) (bool, error) {
+	// Check entries for nil first.
+	if entry == nil && entry2 != nil {
+		return true, nil
+	} else if entry != nil && entry2 == nil {
+		return false, nil
+	}
```
lukechampine, line 105: this doesn't catch `entry == nil && entry2 == nil`. Maybe rewrite as:
```go
if entry2 == nil {
    // A nil entry never replaces an existing entry.
    return false, nil
} else if entry == nil {
    // A non-nil entry always replaces a nil entry.
    return true, nil
}
```

modules/registry.go
```diff
@@ -79,6 +93,41 @@ func (entry RegistryValue) Sign(sk crypto.SecretKey) SignedRegistryValue {
...
+	return false, ErrSameRevNum
+}
+
+// IsRegistryEntryExistErr returns true if the provided error is related to the
+// host already storing a higher priority registry entry.
+func IsRegistryEntryExistErr(err error) bool {
+	if errors.Contains(err, ErrLowerRevNum) || errors.Contains(err, ErrInsufficientWork) || errors.Contains(err, ErrSameRevNum) {
+		return true
+	}
+	return false
```
lukechampine, line 128: `return errors.Contains...`?

modules/host/contractmanager/sectorupdate_test.go
```diff
@@ -1962,7 +1962,7 @@ func TestAddVirtualSectorOverflow(t *testing.T) {
 		t.SkipNow()
 	}
 	t.Parallel()
-	cmt, err := newContractManagerTester("TestAddVirtualSectorMassiveParallel")
+	cmt, err := newContractManagerTester("TestAddVirtualSectorOverflow")
```
DavidVorick, line 1965: t.Name()

PR #11, opened 2021-05-27
Upgrade `log` dependency

peterjan:
This PR upgrades the log dependency to include the recent changes (https://gitlab.com/NebulousLabs/log/-/merge_requests/10). These changes should be a perfect drop-in, only adding functionality that we require in `skyd`. Because `skyd` still imports some interfaces that expose the `persist.Logger` from `siad`, we need `siad` to upgrade and release a new version.


lukechampine:
Shouldn't [Minimal Version Selection](https://research.swtch.com/vgo-mvs) take care of this? If you run:
```
go get gitlab.com/NebulousLabs/log@v0.0.0-20210525183242-36461cfd56ab
```
in the `skyd` repo, it should update your `go.mod` to use that version, meaning that any dependencies will use it as well.

PR #19, opened 2021-07-01
Improve Performance of Some Tests in `siatest/renter`

chris124567:
This pull request improves the performance of various tests in `siatest/renter`.  Many tests in this directory have large initialization times due to having to initialize many test hosts.  This pull request lowers the amount of test hosts created in tests where it was unnecessarily high which improves performance in these tests by about 20-30%.


chris124567:
Note: None of the tests that failed appear to have anything to do with the tests I modified.

PR #22, opened 2021-07-21
Improve contract lifecycle logging

n8maninger:
Adds context to the host log for contract actions


modules/host/storageobligations.go
```diff
@@ -1158,7 +1158,7 @@ func (h *Host) threadedHandleActionItem(soid types.FileContractID) {
 	})
 	h.mu.RUnlock()
 	if err != nil {
-		h.log.Println("Could not get storage obligation:", err)
+		h.log.Printf("contract %s action: Could not get storage obligation: %s", soid, err)
```
lukechampine, line 1161: I prefer to use `%v` as the default formatting verb. It handles `nil` a bit more gracefully.

PR #23, opened 2021-07-27
add block height to modules.ConsensusChange

### LINKED ISSUE #24: Wallet height exceeds consensus height
# BUG REPORT
When adding watch addresses while the wallet is still scanning the wallet's block height will exceed the actual block height. This also created an issue with "invalid signatures" for a few hosts right before the Sia Foundation hard fork. 

Reinitializing the wallet by deleting the `wallet` folder and recovering from seed gets everything back in sync.

## Expected Behavior
All modules should report the same height

## Observed Behavior
+ `siac consensus -v` reports the correct height and block ID
+ `siac wallet` reports height above 300000

## How to reproduce it (as minimally and precisely as possible)
1. Start a synced Sia node with no wallet
2. Generate a new wallet seed
3. Add a watch address 
4. Wait for the wallet to fully scan
5. Check the wallet's height

## Environment
* Sia version: v1.5.6

n8maninger:
Adds `BlockHeight` to `modules.ConsensusChange` and updates all existing consensus subscribers to use it.

## Explorer
Only needs to calculate the height before changes are applied and increments it so each block added to the explorer corresponds to the correct height.

## Host
Calculates the height before blocks are applied then increments it for each applied block so that action items are properly handled.

### Account manager
Changes `AccountManager.callConsensusChanged` to only pass the old height and use `ConsensusChange.BlockHeight` for the new height.

## Miner
Uses `ConsensusChange.Blockheight` directly, just needs to know the height to calculate the correct block subsidy.

### Contractor
Calculates the height before blocks are applied and increments it for each applied block so `Contractor.findRecoverableContracts` can set start height.

### Recovery scanner
Changes `atomicRecoveryScanHeight` to be set directly to `cc.BlockHeight` instead of calculated. It's only used to print the recovery status.

### Host DB
Sets the host db's block height to `cc.BlockHeight` before checking for host announcements to preserve its existing behavior.

## Transaction pool
Sets `tp.blockHeight` after reverting and applying the blocks since the loops are not dependent on `tp.blockHeight`.

## Wallet
Calculates the height before blocks are applied in `applyHistory`, then sets the height once in `ProcessConsensusChange`. Removes the need for `dbGetConsensusHeight` in `applyHistory`, `revertHistory`, and `ProcessConsensusChange`.

Closes #24

### DIFF
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -93,6 +93,10 @@ type (
 		// and applied blocks.
 		ID ConsensusChangeID
 
+		// BlockHeight is the height of the chain after all blocks included in
+		// this change have been reverted and applied.
+		BlockHeight types.BlockHeight
+
 		// RevertedBlocks is the list of blocks that were reverted by the change.
 		// The reverted blocks were always all reverted before the applied blocks
 		// were applied. The revered blocks are presented in the order that they
--- a/modules/consensus.go
+++ b/modules/consensus.go
@@ -272,6 +272,14 @@ func (cc *ConsensusChange) AppendDiffs(diffs ConsensusChangeDiffs) {
 	cc.SiafundPoolDiffs = append(cc.SiafundPoolDiffs, diffs.SiafundPoolDiffs...)
 }
 
+// InitialHeight returns the height of the consensus before blocks are applied.
+func (cc *ConsensusChange) InitialHeight() types.BlockHeight {
+	if cc.BlockHeight == 0 {
+		return 0
+	}
+	return cc.BlockHeight - types.BlockHeight(len(cc.AppliedBlocks))
+}
+
 // MarshalSia implements encoding.SiaMarshaler.
 func (cc ConsensusChange) MarshalSia(w io.Writer) error {
 	return encoding.NewEncoder(w).EncodeAll(
--- a/modules/miner/update.go
+++ b/modules/miner/update.go
@@ -256,32 +256,7 @@ func (m *Miner) ProcessConsensusChange(cc modules.ConsensusChange) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	// Update the miner's understanding of the block height.
-	for _, block := range cc.RevertedBlocks {
-		// Only doing the block check if the height is above zero saves hashing
-		// and saves a nontrivial amount of time during IBD.
-		if m.persist.Height > 0 || block.ID() != types.GenesisID {
-			m.persist.Height--
-		} else if m.persist.Height != 0 {
-			// Sanity check - if the current block is the genesis block, the
-			// miner height should be set to zero.
-			m.log.Critical("Miner has detected a genesis block, but the height of the miner is set to ", m.persist.Height)
-			m.persist.Height = 0
-		}
-	}
-	for _, block := range cc.AppliedBlocks {
-		// Only doing the block check if the height is above zero saves hashing
-		// and saves a nontrivial amount of time during IBD.
-		if m.persist.Height > 0 || block.ID() != types.GenesisID {
-			m.persist.Height++
-		} else if m.persist.Height != 0 {
-			// Sanity check - if the current block is the genesis block, the
-			// miner height should be set to zero.
-			m.log.Critical("Miner has detected a genesis block, but the height of the miner is set to ", m.persist.Height)
-			m.persist.Height = 0
-		}
-	}
-
+	m.persist.Height = cc.BlockHeight
 	// Update the unsolved block.
 	m.persist.UnsolvedBlock.ParentID = cc.AppliedBlocks[len(cc.AppliedBlocks)-1].ID()
 	m.persist.Target = cc.ChildTarget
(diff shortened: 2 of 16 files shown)


modules/miner/update.go
```diff
@@ -256,32 +256,7 @@ func (m *Miner) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-			m.persist.Height++
-		} else if m.persist.Height != 0 {
-			// Sanity check - if the current block is the genesis block, the
-			// miner height should be set to zero.
-			m.log.Critical("Miner has detected a genesis block, but the height of the miner is set to ", m.persist.Height)
-			m.persist.Height = 0
-		}
-	}
-
+	m.persist.Height = cc.BlockHeight
```
lukechampine, line 259: lol what an insane diff

modules/renter/contractor/update.go
```diff
@@ -90,12 +90,17 @@ func (c *Contractor) ProcessConsensusChange(cc modules.ConsensusChange) {
...
 		// Remove recoverable contracts found in reverted block.
 		c.removeRecoverableContracts(block)
 	}
+
+	// Calculate the height before the blocks were applied. If this is the
+	// genesis change set the height to 0.
+	c.blockHeight = cc.BlockHeight - types.BlockHeight(len(cc.AppliedBlocks))
+	if cc.BlockHeight == 0 {
+		c.blockHeight = 0
+	}
```
lukechampine, line 102: Maybe add a `ConsensusChange.InitialHeight()` method to simplify this? Or `PreChangeHeight` and `PostChangeHeight` fields? 🤔

modules/consensus.go
```diff
@@ -268,6 +272,14 @@ func (cc *ConsensusChange) AppendDiffs(diffs ConsensusChangeDiffs) {
 	cc.SiafundPoolDiffs = append(cc.SiafundPoolDiffs, diffs.SiafundPoolDiffs...)
 }
 
+// InitialHeight returns the height of the consensus before blocks are applied.
+func (cc *ConsensusChange) InitialHeight() types.BlockHeight {
+	if cc.BlockHeight == 0 {
+		return 0
+	}
+	return cc.BlockHeight - types.BlockHeight(len(cc.AppliedBlocks))
```
lukechampine, line 280: doesn't this need to account for `cc.RevertedBlocks` as well?

modules/renter/hostdb/update.go
```diff
@@ -90,31 +90,9 @@ func (hdb *HostDB) ProcessConsensusChange(cc modules.ConsensusChange) {
...
-		} else if hdb.blockHeight != 0 {
-			// Sanity check - if the current block is the genesis block, the
-			// hostdb height should be set to zero.
-			hdb.staticLog.Critical("Hostdb has detected a genesis block, but the height of the hostdb is set to ", hdb.blockHeight)
-			hdb.blockHeight = 0
-		}
-	}
+	// Set the block height before applying blocks to preserve previous
+	// behavior.
+	hdb.blockHeight = cc.BlockHeight
```
lukechampine, line 95: is this a bug?

modules/wallet/update.go
```diff
@@ -546,6 +527,10 @@ func (w *Wallet) ProcessConsensusChange(cc modules.ConsensusChange) {
 		w.log.Severe("ERROR: failed to update consensus change ID:", err)
 		w.dbRollback = true
 	}
+	if err := dbPutConsensusHeight(w.dbTx, cc.BlockHeight); err != nil {
```
lukechampine, line 530: do we still need this DB value at all?

modules/consensus.go
n8maninger: It doesn't need to, but it may make more sense if it did. The height is expected to be after reverting and before applying. If the subscriber doesn't use `cc.BlockHeight` directly it follows this pattern:
```go
for _, b := range cc.RevertedBlocks {
  // Revert block
}

height := cc.InitialHeight()
for _, b := range cc.AppliedBlocks {
  // Apply block
  height++
}
```

modules/renter/hostdb/update.go
n8maninger: Yes, the `FirstSeen` height in the hostdb isn't necessarily the height of the block the announcement was seen in.

modules/wallet/update.go
n8maninger: We could refactor it out and load it from consensus initially, but the wallet still pulls it from the database with `dbGetConsensusHeight` in a few places.

modules/renter/hostdb/update.go
lukechampine: Ok. Not a big deal I guess. (In practice, 99% of ConsensusChanges are single-block applications.)

modules/consensus.go
lukechampine: Ah, I see. Kinda weird, but if we explain it in the docstring then it should be fine.

modules/renter/hostdb/update.go
lukechampine: ok, not a big deal I guess

PR #32, opened 2021-08-15
renter: fix zero stream upload

ivandeex:
The renter module can accept zero size files (see [upload.go](https://github.com/SiaFoundation/siad/blob/v1.5.6/modules/renter/upload.go#L104)).

However the upload streamer will fail for an empty stream with
```
unable to stream an upload from a reader; upload streamer failed to get all data available
```
This makes a number of unit tests fail for the Sia [rclone](https://github.com/rclone/rclone) backend developed by @MSevey at https://github.com/rclone/rclone/pull/4514.

This pull request adds the appropriate check in [uploadstreamer.go](https://github.com/SiaFoundation/siad/blob/v1.5.6/modules/renter/uploadstreamer.go#L242) to make zero stream uploads work as expected.

@lukechampine Please review

### DIFF
--- a/modules/renter/uploadstreamer.go
+++ b/modules/renter/uploadstreamer.go
@@ -218,6 +218,15 @@ func (r *Renter) callUploadStreamFromReader(up modules.FileUploadParams, reader 
 		}
 	}()
 
+	// Check if stream has at least one byte. No need to upload empty data.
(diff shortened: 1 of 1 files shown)


lukechampine:
Thanks for the fix!

PR #35, opened 2021-08-28
Fix release scripts

n8maninger:
Changes the `release-scripts/release.sh` to use the makefile instead of having to update `build.NodeVersion` in multiple places. Also removes `build.ReleaseTag` since it seems unnecessary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
 # These variables get inserted into ./build/commit.go
-BUILD_TIME=$(shell date)
 GIT_REVISION=$(shell git rev-parse --short HEAD)
 GIT_DIRTY=$(shell git diff-index --quiet HEAD -- || echo "✗-")
+ifeq ("$(GIT_DIRTY)", "✗-")
+	BUILD_TIME=$(shell date)
+else
+	BUILD_TIME=$(shell git show -s --format=%ci HEAD)
+endif
 
 ldflags= \
 -X "go.sia.tech/siad/build.BinaryName=siad" \
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ dev-race:
 	GORACE='$(racevars)' go install -race -tags='dev debug profile netgo' -ldflags='$(ldflags)' $(pkgs)
 
 static:
-	go build -o release/ -tags='netgo' -ldflags='-s -w $(ldflags)' $(release-pkgs)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -1,7 +1,11 @@
 # These variables get inserted into ./build/commit.go
-BUILD_TIME=$(shell date)
 GIT_REVISION=$(shell git rev-parse --short HEAD)
 GIT_DIRTY=$(shell git diff-index --quiet HEAD -- || echo "✗-")
+ifeq ("$(GIT_DIRTY)", "✗-")
+	BUILD_TIME=$(shell date)
+else
+	BUILD_TIME=$(shell git show -s --format=%ci HEAD)
```
lukechampine, line 7: Does this set the build time to the last commit timestamp? Is that desirable?

release-scripts/release.sh
```diff
@@ -9,32 +9,41 @@ if [[ -z $version ]]; then
...
+	GOOS=$1 GOARCH=$2 make static
+	mv release/* $folder
+
 	# compile and hash binaries
 	for pkg in siac siad; do
 		bin=$pkg
 		if [ "$os" == "windows" ]; then
 			bin=${pkg}.exe
 		fi
-		GOOS=${os} GOARCH=${arch} go build -a -tags 'netgo' -trimpath -ldflags="$ldflags" -o $folder/$bin ./cmd/$pkg
```
lukechampine, line 30: we'll want to preserve the `-trimpath`

lukechampine:
I'm not sure we should remove `ReleaseTag`. `build.NodeVersion` doesn't contain RC information, so if you run `siad version` on an RC, it won't indicate that you're running an RC.

n8maninger:
> I'm not sure we should remove `ReleaseTag`. `build.NodeVersion` doesn't contain RC information, so if you run `siad version` on an RC, it won't indicate that you're running an RC.

My thought was build.NodeVersion could contain RC information without any issues now since the host and gateway versions are now separate.

Makefile
n8maninger: It’s what release.sh did, I think it was for more reproducible binaries.

lukechampine:
Oh yeah. It looks like we do still call `VersionCmp` on `NodeVersion` in one place (auto-updating). If we take care of that, then we can start adding `-rc1` to `NodeVersion` for RCs.

lukechampine:
hmm, looks like this broke docker: https://github.com/SiaFoundation/siad/actions/runs/1183351061

PR #36, opened 2021-08-29
Remove panic in AccountID.PK()

n8maninger:
Replaces a panic with an error when the ephemeral account id is not a valid length to convert to a public key.

### DIFF
--- a/modules/payment_test.go
+++ b/modules/payment_test.go
@@ -14,7 +14,7 @@ import (
 // TestNewAccountID tests the NewAccountID utility method.
 func TestNewAccountID(t *testing.T) {
 	t.Parallel()
-	aid, sk := NewAccountID()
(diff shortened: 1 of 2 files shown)


modules/payment_test.go
```diff
@@ -183,3 +160,105 @@ func TestAccountIDCompatSiaMarhsal(t *testing.T) {
...
+			false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			fingerprint := crypto.HashAll(tt.message)
+			sig := crypto.SignHash(fingerprint, tt.sk)
+			err := tt.message.Validate(tt.current, tt.current+20, fingerprint, sig)
+			t.Log(err)
```
lukechampine, line 258: leftover?

PR #44, opened 2021-10-02
gateway: Add flag to disable UPnP

### LINKED ISSUE #42: [Feature Request] Switch to disable uPnP discovery
Please can a switch be added to disable uPnP discovery and port mapping?

My use case is that I have a /29 block of IP addresses (x.y.z.88-95). My router base address is x.y.z.90 and I have allocated Sia x.y.z.91 using a 1-to-1 NAT function. In the router, uPnP mappings take precedence over 1-to-1 NAT so siad discovers its external IP as x.y.z.90 and not x.y.z.91. Turning off uPnP in the router is a global setting. I have had to resort to blocking incoming TCP:1900 (uPnP discovery) from the Sia LAN IP to disable the uPnP discovery from Sia, but it would be much better if siad had a switch to stop it.

lukechampine:
Fixes #42

### DIFF
--- a/cmd/siad/main.go
+++ b/cmd/siad/main.go
@@ -41,6 +41,7 @@ type Config struct {
 
 		Modules           string
 		NoBootstrap       bool
+		UseUPNP           bool
 		RequiredUserAgent string
 		AuthenticateAPI   bool
 		TempPassword      bool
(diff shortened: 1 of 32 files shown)


cmd/siad/main.go
```diff
@@ -194,6 +195,7 @@ func main() {
 	root.Flags().StringVarP(&globalConfig.Siad.APIaddr, "api-addr", "", "localhost:9980", "which host:port the API server listens on")
 	root.Flags().StringVarP(&globalConfig.Siad.SiaDir, "sia-directory", "d", "", "location of the sia directory")
 	root.Flags().BoolVarP(&globalConfig.Siad.NoBootstrap, "no-bootstrap", "", false, "disable bootstrapping on this run")
+	root.Flags().BoolVarP(&globalConfig.Siad.UseUPNP, "upnp", "", true, "use UPnP for port forwarding and external IP discovery")
```
n8maninger, line 198: I also prefer it disabled by default, but should this default to true? There could be users relying on the existing functionality and unable to change it when using Sia-UI.

lukechampine: the default is `true`, so UPnP will remain enabled by default

n8maninger: Oh you’re right, bad comprehension on my part 🤦‍♂️

PR #50, opened 2021-10-29
Introduce sectorMu in Host to reduce download latency

ChrisSchinnerl:
This PR introduces a new mutex in the contract manager. The mutex itself comes with a description in the docstring but I'll summarize it here as well.

The host syncs all of its persistence to disk every 500ms. It does so while holding the wallet mutex. That same mutex is required to read sector locations, storage folders and lock storage locations in memory.

That's an issue because the persistence can take hundreds of milliseconds or as I noticed while running a local test even a few seconds. That was on an SSD. I assume most hosts have HDDs which take even longer.

So when a user downloads a sector, they have to acquire that lock to locate the sector which potentially slows down the TTFB by a lot.

This is why I added a dedicated log for the fields required for downloads which consistently reduces the lookup time to <1ms times since they no longer require waiting for syncing data to disk.

NOTE: I understand that the host is a gentle creature and the locking in the host is pretty magical so I went for this rather hacky but not very invasive solution. Tbh the host deserves a full overhaul but I don't think anybody got the capacity for that at the moment.

### DIFF
--- a/modules/host/contractmanager/sector.go
+++ b/modules/host/contractmanager/sector.go
@@ -3,8 +3,10 @@ package contractmanager
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"go.sia.tech/siad/build"
 	"go.sia.tech/siad/crypto"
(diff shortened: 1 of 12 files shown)


modules/host/contractmanager/sector.go
```diff
@@ -127,14 +129,16 @@ func (cm *ContractManager) ReadPartialSector(root crypto.Hash, offset, length ui
...
 	defer cm.tg.Done()
 	id := cm.managedSectorID(root)
-	cm.wal.managedLockSector(id)
-	defer cm.wal.managedUnlockSector(id)
 
 	// Fetch the sector metadata.
-	cm.wal.mu.Lock()
+	start := time.Now()
+	cm.sectorMu.Lock()
+	fmt.Println("locked", time.Since(start))
```
lukechampine, line 136: debug println left in?

modules/host/contractmanager/sector.go
```diff
@@ -170,16 +174,32 @@ func (cm *ContractManager) HasSector(sectorRoot crypto.Hash) bool {
...
+		}
+		wal.cm.lockedSectors[id] = sl
+	}
+
+	// Block until the sector is available.
+	sl.mu.Lock()
+}
+
 // managedLockSector grabs a sector lock.
 func (wal *writeAheadLog) managedLockSector(id sectorID) {
```
lukechampine, line 201: is this function still necessary? Should it call `lockSector`?

lukechampine:
Seems like a worthwhile change to me.
Re: host overhaul, I expect the Sia v2 host will be based on [`us/host`](https://github.com/lukechampine/us/tree/master/host), augmented to support RHPv3.

modules/host/contractmanager/sector.go
ChrisSchinnerl: removed

modules/host/contractmanager/sector.go
ChrisSchinnerl: changed it to call `lockSector`.

PR #53, opened 2021-11-04
Fix sectorMu deadlocks

### LINKED ISSUE #52: Host sectorMu deadlock
@lukechampine @ChrisSchinnerl #50 introduced a deadlock on startup: 

https://github.com/SiaFoundation/siad/blob/9846fccdd0ee2253c7c4e0b2ac73823917f46c4f/modules/host/contractmanager/contractmanager.go#L229

https://github.com/SiaFoundation/siad/blob/9846fccdd0ee2253c7c4e0b2ac73823917f46c4f/modules/host/contractmanager/persist.go#L165 

Does it make sense to remove the lock from `loadSectorLocations` and refactor `threadedFolderRecheck`?

n8maninger:
This fixes the two issues I noticed in #50

+ Fixes a deadlock on startup when calling `loadSectorLocations` from `newContractManager`.
+ Fixes a deadlock in `managedLockSector` when waiting to acquire a sector lock.
+ Remove usage of `lockSector` to prevent holding `sectorMu` while potentially waiting on it in `managedUnlockSector`

Closes #52

### DIFF
--- a/modules/host/contractmanager/storagefolder.go
+++ b/modules/host/contractmanager/storagefolder.go
@@ -315,30 +315,32 @@ func (cm *ContractManager) threadedFolderRecheck() {
 		// to the contract manager.
 		cm.wal.mu.Lock()
 		cm.sectorMu.Lock()
-		sfs := make([]*storageFolder, 0, len(cm.storageFolders))
+		var unavailable []*storageFolder
 		for _, sf := range cm.storageFolders {
-			sfs = append(sfs, sf)
+			if atomic.LoadUint64(&sf.atomicUnavailable) == 1 {
+				unavailable = append(unavailable, sf)
+			}
 		}
 		cm.sectorMu.Unlock()
 
-		for _, sf := range sfs {
-			if atomic.LoadUint64(&sf.atomicUnavailable) == 1 {
-				var err1, err2 error
-				sf.metadataFile, err1 = cm.dependencies.OpenFile(filepath.Join(sf.path, metadataFile), os.O_RDWR, 0700)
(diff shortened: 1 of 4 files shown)


lukechampine:
Looks alright to me, but I'd feel more confident if @ChrisSchinnerl review this as well.

modules/host/contractmanager/storagefolder.go
```diff
@@ -315,13 +315,8 @@ func (cm *ContractManager) threadedFolderRecheck() {
...
 		cm.wal.mu.Lock()
 		cm.sectorMu.Lock()
-		sfs := make([]*storageFolder, 0, len(cm.storageFolders))
-		for _, sf := range cm.storageFolders {
-			sfs = append(sfs, sf)
-		}
-		cm.sectorMu.Unlock()
 
-		for _, sf := range sfs {
+		for _, sf := range cm.storageFolders {
```
ChrisSchinnerl, line 319: Did this cause issues as well? The reason I copied the folders first was to avoid opening a file while holding `sectorMu` every second we trigger the recheck.

But I guess if we only do this once per storage folder and the remaining times we just check the atomic and skip opening the file it's fine.

ChrisSchinnerl (approved):

modules/host/contractmanager/storagefolder.go
n8maninger: It wasn't causing problems, but it seemed like the safer route after removing `sectorMu` from `loadSectorLocations`. I'll put the copy back and also wrap `loadSectorLocations`.

n8maninger: Do the changes look okay with what you were going for?

n8maninger:
@lukechampine look good? Looks like only the flaky renter tests failed.

PR #54, opened 2021-11-05
Optimise heap allocations for "generateSpendableKey"

ChrisSchinnerl:
While I was benchmarking skyd on one of our development servers, I noticed that 30%+ of all allocated objects were created in `generateSpendableKey`. We are talking about tens of millions of objects.

This MR is something I cooked up this morning to get rid of those allocations and it worked. Essentially I'm turning `spendableKey` into a value type by introducing a new `UnlockCondition` type which is a specific version of `UnlockConditions` containing only one key.

That way we only use value fields and arrays instead of keys with arbitrary slices.

Let me know what you think. There might be other ways that are less intrusive but this seemed like a quick low-hanging fruit.

### DIFF
--- a/cmd/siac/daemoncmd.go
+++ b/cmd/siac/daemoncmd.go
@@ -93,6 +93,8 @@ directory in the siad data directory.`,
 // alertscmd prints the alerts from the daemon. This will not print critical
 // alerts as critical alerts are printed on every siac command
 func alertscmd() {
+	const maxAlerts = 1000
+
 	al, err := httpClient.DaemonAlertsGet()
 	if err != nil {
 		fmt.Println("Could not get daemon alerts:", err)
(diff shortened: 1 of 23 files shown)


lukechampine:
My first reaction was "using value types means that secret keys will be copied around in memory, which is bad." My second reaction was "oh wait `crypto.SecretKey` is a value type so I guess we already do that *everywhere*." This will be definitely rectified in v2. (`core` does not have its own secret key type; we just use `ed25519.PrivateKey`.)

My third reaction is that we probably don't need a new `types.UnlockCondition` type. Instead, we can just inline the `Timelock` and `PublicKey` fields into `wallet.spendableKey`. (And while `types.UnlockCondition` is a cute name, it's a bit of a misnomer since technically there are *two* conditions -- the pubkey and the timelock. 😉)

And my fourth reaction is that we can further save allocations by writing a zero-alloc `(spendableKey).UnlockHash()` method, based on the `us` [`wallet.StandardAddress`](https://github.com/lukechampine/us/blob/7d274345277846d9132ce858701dfa875c1a879a/wallet/wallet.go#L77) function.

lukechampine:
There's a failing wallet test, but otherwise this looks good. Are you able to benchmark again with this code and observe a reduction in allocations?

lukechampine:
Looks like this test broke:
```
  Error: test go.sia.tech/siad/modules/wallet/TestIntegrationLoad2of3Siag failed in 0.35s:
  === RUN   TestIntegrationLoad2of3Siag
      unseeded_test.go:79: wrong number of pubkeys
      unseeded_test.go:87: expecting a siafund balance of 7000 from the 2of3 key
      unseeded_test.go:93: insufficient balance
  --- FAIL: TestIntegrationLoad2of3Siag (0.35s)
  ```

PR #56, opened 2021-11-24
v2: add new wallet commands and API endpoints

chris124567:
This adds the following wallet subcommands:

    address         display wallet address
    addresses       display all addresses generated by the wallet
    transactions    display transactions relevant to the wallet
    broadcast       broadcast a JSON encoded transaction
    send            send siacoins or siafunds

`send` can only actually send siacoins right now (`send siacoins 123 addr:xxxx`) but I made it be its own command with subcommands so that Siafund support can be easily added once Wallet has Siafund support.  The main thing I wasn't sure about was the validation context the Server should use so I just got the tip from the chain manager (see https://github.com/SiaFoundation/siad/blob/d421a0a28023ac38aaa9f053751331b9abad2f10/cmd/siad/main.go#L126).


api/api.go
```diff
@@ -1,7 +1,26 @@
...
+	Addresses []types.Address `json:"addresses"`
+}
+
+type Transactions struct {
+	Transactions []wallet.Transaction `json:"transactions"`
+}
+
+type SendReceipt struct {
+	Error *error `json:"success"`
+}
```
lukechampine, line 26: For API errors, we use `http.Error` -- see examples in e.g. https://github.com/lukechampine/walrus/blob/master/server.go

api/client.go
```diff
@@ -53,6 +54,35 @@ func (c *Client) WalletBalance() (resp WalletBalance, err error) {
...
+
+func (c *Client) SendSiacoins(out types.SiacoinOutput) (resp SendReceipt, err error) {
+	err = c.post("/wallet/send/siacoins", out, &resp)
+	return
+}
+
+func (c *Client) BroadcastTransaction(txn types.Transaction) (resp SendReceipt, err error) {
+	err = c.post("/wallet/broadcast", txn, nil)
+	return
+}
```
lukechampine, line 84: I wonder if we should "namespace" the client, i.e.
```go
type Client struct {
    Wallet WalletClient
    TxPool TxPoolClient
    // etc.
}
```
Then we could write `c.Wallet.Addresses()` instead of `c.WalletAddresses()`, and we wouldn't need to worry about name collisions. 🤔

api/server.go
```diff
@@ -2,10 +2,13 @@ package api
 
 import (
 	"encoding/json"
+	"github.com/julienschmidt/httprouter"
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/siad/v2/p2p"
+	"go.sia.tech/siad/v2/wallet"
```
lukechampine, line 8: `goimports` should automatically group these with the other non-stdlib packages at the bottom

api/server.go
```diff
@@ -26,10 +29,17 @@ func writeJSON(w http.ResponseWriter, v interface{}) {
...
+	Addresses() []types.Address
+	Transactions() []wallet.Transaction
+	FundTransaction(txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.ElementID, func(), error)
+	SignTransaction(vc consensus.ValidationContext, txn *types.Transaction, toSign []types.ElementID) error
 }
 
 type Server struct {
-	w Wallet
+	w  Wallet
+	s  *p2p.Syncer
```
lukechampine, line 41: as with `Wallet`, we'll want to define a `Syncer` interface that `*p2p.Syncer` satisfies

api/server.go
```diff
@@ -38,12 +48,66 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+	defer writeJSON(w, SendReceipt{
+		Error: &err,
+	})
+
+	var txn types.Transaction
+	if err := json.NewDecoder(req.Body).Decode(&txn); err != nil {
+		return
+	}
+	// TODO: Validate?
+	s.s.BroadcastTransaction(txn, nil)
```
lukechampine, line 95: ah -- in addition to broadcasting the transaction, we also want to add it to our own txpool. That takes care of validation too, since `tp.AddTransaction` will validate the transaction before adding it.

api/server.go
```diff
@@ -38,12 +48,66 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+}
+
 // NewServer returns an HTTP handler that serves the siad API.
-func NewServer(w Wallet) http.Handler {
+func NewServer(w Wallet, sy *p2p.Syncer, vc *consensus.ValidationContext) http.Handler {
```
lukechampine, line 99: maybe use `srv` for `Server` to free up `s` for `Syncer` :)

cmd/siac/main.go
```diff
@@ -36,12 +39,53 @@ Subcommands:
...
+	sendSiacoinsUsage = `Usage:
+siac [flags] wallet [flags] send siacoins [amount] [dest]
+
+Send siacoins to an address.
+`
+
+	broadcastUsage = `Usage:
+siac [flags] wallet [flags] broadcast [txn]
+
+Broadcast a JSON encoded transaction.
```
lukechampine, line 88: 👍🏻

cmd/siac/main.go
```diff
@@ -75,7 +119,14 @@ func main() {
 	walletCmd := flagg.New("wallet", walletUsage)
 	walletBalanceCmd := flagg.New("balance", walletBalanceUsage)
 	exactBalance := walletBalanceCmd.Bool("exact", false, "print balance in Hastings")
-
+	walletAddressCmd := flagg.New("address", addressUsage)
+	walletAddressesCmd := flagg.New("addresses", addressesUsage)
+	walletTransactionsCmd := flagg.New("transactions", transactionsUsage)
+	transactionID := walletTransactionsCmd.String("id", "", "print only a specific transaction by its ID")
```
lukechampine, line 125: seems kinda unnecessary; you can `grep` for a particular ID. Maybe instead we could have a separate `transaction` command that nicely formats all of the details of a particular transaction.

cmd/siac/main.go
```diff
@@ -118,5 +174,88 @@ func main() {
...
+		}
+		transactions, err := getClient().Transactions()
+		check("couldn't get transactions:", err)
+
+		var parsedTxnID types.TransactionID
+		if flagg.IsDefined(cmd, "id") {
+			check("couldn't parse transaction ID:", json.Unmarshal([]byte(`"`+*transactionID+`"`), &parsedTxnID))
+		}
+
+		w := tabwriter.NewWriter(os.Stdout, 8, 8, 0, '\t', 0)
```
lukechampine, line 210: `tabwriter.NewWriter` is one of the worst APIs in the stdlib lol. I can never remember what the args do. It looks like we always used `tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)` in `siac`, though.

lukechampine:
For the wallet-related code in `siac`, we can crib from https://github.com/lukechampine/walrus-cli/blob/master/main.go. Particularly the commands/flags for constructing, signing, and broadcasting txns.

Relatedly, I think the API doesn't need a `send` endpoint. The logic is pretty easy to implement client-side: just fund, sign, and broadcast. (I guess we'll need a `sign` endpoint.)

I wouldn't worry about siafunds for now. We may end up releasing a separate binary for managing siafunds. The old wallet managed both siacoins and siafunds and it was always a bit of a mess.

Lastly, as for the `ValidationContext` -- the reason it's passed to `SignTransaction` is because we need to compute the hash to sign, and the details of that computation are consensus rules, and all consensus rules are tied to `ValidationContext`. So we want to pass the tip context to `SignTransaction` -- and it should be the *current* tip, not just the tip when we start `siad`. That means that the `Server` needs to get the current tip context somehow in order to pass it to `SignTransaction`. Probably `ChainManager` should have a `TipContext` method. Alternatively, `Server` could subscribe to the `ChainManager` and get the current context that way, but that's sort of a weird design ("why should the server have to subscribe to updates from the thing it's serving?").

api/client.go
chris124567: How likely/common are collisions? I guess we'll see as we add more endpoints

chris124567:
Alright now we got `sign` and got rid of `send` among other things.  I made the sign command sign and fund a transaction because separating them seemed unnecessary.

api/server.go
```diff
@@ -38,12 +60,76 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+
+	var amount types.Currency
+	for _, out := range txn.SiacoinOutputs {
+		amount = amount.Add(out.Value)
+	}
+
+	cm := s.s.ChainManager()
+	vc, err := cm.ValidationContext(cm.Tip())
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
n8maninger, line 90: `http.StatusInternalServerError` would be a better fit here.

chris124567:
Cleaned up in some places and added new commands mostly relating to peers and the transaction pool.

api/server.go
```diff
@@ -38,12 +65,123 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+	}
+
+	cm := s.s.ChainManager()
+	vc, err := cm.ValidationContext(cm.Tip())
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if toSign, _, err := s.w.FundTransaction(&txn, amount, nil); err != nil {
```
lukechampine, line 99: an endpoint named "sign" should not also be funding the transaction. In `walrus`, the client requests the wallet's unspent outputs, and uses them to construct the transaction locally. We should take the same approach here. We can add a helper endpoint like `/wallet/fund` later if needed.

api/server.go
```diff
@@ -38,12 +65,123 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+			outflow = outflow.Add(in.Parent.Value)
+		}
+		txns = append(txns, wallet.Transaction{
+			Raw:     txn.DeepCopy(),
+			ID:      txn.ID(),
+			Inflow:  inflow,
+			Outflow: outflow,
+		})
+	}
+	writeJSON(w, Transactions{txns})
```
lukechampine, line 140: txpool routes should probably just return `types.Transaction`, not `wallet.Transaction`, since the majority of transactions in the pool won't be relevant to the wallet.

cmd/siac/main.go
```diff
@@ -86,12 +179,34 @@ func main() {
...
+					{Cmd: syncerAddressCmd},
+					{Cmd: syncerConnectCmd},
 				},
 			},
 		},
 	})
 	args := cmd.Args()
 
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
+	defer w.Flush()
```
lukechampine, line 208: A tabwriter is useful for multiple commands, but it should be initialized separately within each command's scope. We can define a helper function though, so that we don't need to copy-paste the same magic arguments each time:
```go
func makeTabWriter() *tabwriter.Writer {
    return tabwriter.NewWriter(
        os.Stdout,
        0,   // minwidth: zero, since we're left-aligning
        0,   // tabwidth: zero, since we're using ' ',  not '\t'
        2,   // padding:  two, so that columns aren't right next to each other
        ' ', // padchar:  spaces, not tabs
        0,   // flags:    none
    )
}
```

cmd/siac/main.go
```diff
@@ -118,5 +233,113 @@ func main() {
...
+		}
+
+	case walletSignCmd:
+		if len(args) != 1 {
+			cmd.Usage()
+			return
+		}
+
+		var txn types.Transaction
+		check("failed to decode JSON:", json.Unmarshal([]byte([]byte(args[0])), &txn))
```
lukechampine, line 277: redundant `[]byte` here

We'll also want to support reading/writing these transactions from disk

wallet/wallet.go
```diff
@@ -107,6 +107,17 @@ func (w *HotWallet) NextAddress() types.Address {
...
+// Adresses returns all addresses generated by the wallet.
+func (w *HotWallet) Addresses() []types.Address {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	a := make([]types.Address, w.store.SeedIndex())
+	for i := uint64(0); i < w.store.SeedIndex(); i++ {
+		a[i] = types.StandardAddress(w.seed.PublicKey(i))
+	}
+	return a
+}
```
lukechampine, line 119: This might be premature optimization, but technically only the `SeedIndex` needs to be covered by the mutex (and deriving addresses isn't the fastest operation, so we should avoid doing it under lock if possible).

```go
w.mu.Lock()
addrs := make([]types.Address, w.store.SeedIndex())
w.mu.Unlock()
for i := range addrs {
    addrs[i] = types.StandardAddress(w.seed.PublicKey(uint64(i)))
}
return addrs
```

api/client.go
lukechampine: It's nice because `Wallet.Transactions()` maps cleanly to the corresponding `/wallet/transactions` endpoint and `siac wallet transactions` command.

The main downside I can think of is that it means you can't define a single interface with all of the Client methods; instead, you'd have to do something like
```go
type APIClient interface {
    WalletAddress() (api.WalletAddress, error)
    WalletTransactions() (api.Transactions, error)
    TxpoolTransactions() (api.Transactions, error)
}

type clientWrapper struct {
    c *api.Client
}
func (cw clientWrapper) WalletAddress() (api.WalletAddress, error) {
    return cw.c.Wallet.Address()
}
func (cw clientWrapper) WalletTransactions() (api.Transactions, error) {
    return cw.c.Wallet.Transactions()
}
func (cw clientWrapper) TxpoolTransactions() (api.Transactions, error) {
    return cw.c.Txpool.Transactions()
}
```

You *could*, however, define interfaces for just `Wallet`, or just `Txpool`, etc. So this is only a problem if you're trying to mock the client as a whole.

Let's leave it for now and experiment with this design once the API is more fleshed out.

api/server.go
```diff
@@ -38,12 +65,123 @@ func (s *Server) walletBalanceHandler(w http.ResponseWriter, req *http.Request,
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	} else if err := s.w.SignTransaction(vc, &txn, toSign); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	writeJSON(w, WalletTransaction{txn})
+}
+
+func (s *Server) walletBroadcastTransactionHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
lukechampine, line 109: Broadcasting is a function of the txpool; nothing prevents you from broadcasting transactions that aren't relevant to your own wallet.

p2p/syncer.go
```diff
@@ -612,6 +612,10 @@ func (s *Syncer) Close() error {
 	return s.l.Close()
 }
 
+func (s *Syncer) ChainManager() *chain.Manager {
+	return s.cm
+}
```
lukechampine, line 617: Instead, add a `cm *chain.Manager` field to `api.Server`

n8maninger:
We should consider adding `offset` and `limit` to the transactions interface to paginate the `/wallet/transactions` endpoint. v1 has `startheight` and `endheight` params, but the last `n` transactions would be much more helpful.

lukechampine:
Agreed -- pagination is something that I've always wanted to add. The tricky thing is dealing with races between the API client and the blockchain. Consider that a reorg might occur between two API requests, possibly invalidating some transactions, possibly reordering others, possibly adding new ones.

I can't see an easy way of dealing with this. If we just translate API queries to "database queries", then a reorg can surely break pagination. Alternatively, we could create and retain the full API response in memory, which would preserve its consistency -- but even ignoring the memory overhead in the *best* case scenario, what happens if the client goes away before consuming the full response? Do we attach a `validUntil` field to the paginated result, and discard the memory after that? Open to ideas here.

n8maninger:
I don't think the daemon needs to handle consistency between requests. Pagination will similarly break if new transactions are added between API requests. The consumer should be expected to manage changes to consensus invalidating its data.

lukechampine:
I suppose as a best-effort approach, we could notify the client when their results had been invalidated by a reorg. Using `?since=[timestamp]` as a parameter could also help in the presence of new transactions.

chris124567:
There's everything except pagination.  I guess that should be next.

cmd/siac/main.go
```diff
@@ -118,5 +242,133 @@ func main() {
...
+		cmd.Usage()
+
+	case txpoolBroadcastCmd:
+		if len(args) != 1 {
+			cmd.Usage()
+			return
+		}
+
+		var txn types.Transaction
+		if err := json.Unmarshal([]byte(args[0]), &txn); err != nil {
```
lukechampine, line 320: on further thought, it's unlikely that anyone will be passing a big JSON object as a CLI argument. Let's just require files everywhere.  You can copy the `readTxn` and `writeTxn` helpers from `walrus-cli`: https://github.com/lukechampine/walrus-cli/blob/0f2399a1163f695082275b83aff9974385c71739/main.go#L147-L161

cmd/siad/main.go
```diff
@@ -123,7 +123,7 @@ func main() {
 	}
 	log.Println("api: Listening on", l.Addr())
 	go func() {
-		if err := http.Serve(l, api.NewServer(n.w)); err != nil {
+		if err := http.Serve(l, api.NewServer(n.w, n.s, n.s.ChainManager(), n.tp)); err != nil {
```
lukechampine, line 126: can just pass `n.c`

p2p/syncer.go
lukechampine: (pls remove this method)

api/server.go
```diff
@@ -26,10 +33,30 @@ func writeJSON(w http.ResponseWriter, v interface{}) {
...
+type TransactionPool interface {
+	Transactions() []types.Transaction
+	AddTransaction(txn types.Transaction) error
 }
 
 type Server struct {
-	w Wallet
+	w  Wallet
+	s  Syncer
+	cm *chain.Manager
```
lukechampine, line 58: should define a `ChainManager` interface

api/api.go
```diff
@@ -1,7 +1,45 @@
...
+	Transactions []wallet.Transaction `json:"transactions"`
+}
+
+type TxpoolTransactions struct {
+	Transactions []types.Transaction `json:"transactions"`
+}
+
+type SyncerPeers struct {
+	Peers []gateway.Header `json:"peers"`
+}
```
lukechampine, line 41: usually we prefer to use structs as responses, since you can add more fields to them later without breaking compatibility. But for simple list-type responses like transactions, peers, etc., I think it's fine to use a slice. In other words:
```go
type WalletAddresses []types.Address
type WalletTransactions []wallet.Transaction
type SyncerPeers []gateway.Header
// etc.
```

PR #58, opened 2021-12-03
v2: Add offset and limit to wallet transactions and addresses endpoints

chris124567:
This adds the ability to filter by offset/limit to the wallet transactions and addresses endpoints.  Also performs sanity checks to ensure we do not access out of bounds elements in the slices.

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"net/http"
 	"reflect"
+	"strconv"
+	"time"
 
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/core/net/gateway"
@@ -33,7 +35,7 @@ type Wallet interface {
 	Balance() types.Currency
 	NextAddress() types.Address
 	Addresses() []types.Address
-	Transactions() []wallet.Transaction
+	Transactions(since time.Time, max int) []wallet.Transaction
 	FundTransaction(txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.ElementID, func(), error)
 	SignTransaction(vc consensus.ValidationContext, txn *types.Transaction, toSign []types.ElementID) error
 }
@@ -76,11 +78,49 @@ func (s *Server) walletAddressHandler(w http.ResponseWriter, req *http.Request, 
 }
 
 func (s *Server) walletAddressesHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	writeJSON(w, WalletAddresses(s.w.Addresses()))
+	addresses := s.w.Addresses()
+	start, end := 0, len(addresses)
+	if v := req.FormValue("start"); v != "" {
+		i, err := strconv.Atoi(v)
+		if err != nil || i < 0 || i > len(addresses) {
+			http.Error(w, "invalid start value", http.StatusBadRequest)
+			return
+		}
+		start = i
+	}
+	if v := req.FormValue("end"); v != "" {
+		i, err := strconv.Atoi(v)
+		if err != nil || i < 0 || i < start {
+			http.Error(w, "invalid end value", http.StatusBadRequest)
+			return
+		} else if i > len(addresses) {
+			i = len(addresses)
+		}
+		end = i
+	}
+	writeJSON(w, WalletAddresses(addresses[start:end]))
 }
 
 func (s *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	writeJSON(w, s.w.Transactions())
+	var since time.Time
+	if v := req.FormValue("since"); v != "" {
+		t, err := time.Parse(time.RFC3339, v)
+		if err != nil {
+			http.Error(w, "invalid since value: "+err.Error(), http.StatusBadRequest)
+			return
+		}
+		since = t
+	}
+	max := -1
+	if v := req.FormValue("max"); v != "" {
+		t, err := strconv.Atoi(v)
+		if err != nil {
+			http.Error(w, "invalid max value: "+err.Error(), http.StatusBadRequest)
+			return
+		}
+		max = t
+	}
+	writeJSON(w, s.w.Transactions(since, max))
 }
 
 func (s *Server) walletSignHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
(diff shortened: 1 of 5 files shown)


n8maninger:
Would this be better to handle in the interface over the endpoint? You wouldn't have to load the whole store to get a partial data set.

chris124567:
We could.  That would complicate the wallet and store interfaces a bit though (we'd need something like `Transactions(begin, end int)` and `TransactionsLen()` to use in bounds checking).

lukechampine:
We definitely can't read the whole transaction history into memory for every request, even if we *know* the client wants all of the transactions. I suggest a signature of:
```go
Transactions(since time.Time, max int) ([]wallet.Transaction, error)
```
This should work pretty well. Later, when we add HTTP consensus subscription, clients will be able to detect when their transaction queries are invalidated by a reorg.

chris124567:
OK for the wallet transactions endpoint you now simply specify `since` and `max`.  I used GET query parameters although we have been using POST for requests where the client submits data since in my experience that is the standard practice to specify these kind of limits.  Do we want to do something similar for /wallet/addresses? There's not really a `since` date for addresses but I mean in terms of not loading the full array in memory (you'd have to have a massive list of addresses for this to become a problem).  Also, I could have used binary search when searching for transactions by date to find the first transaction that is from a date `>= since` because the wallet transactions list is sorted but I didn't do that because it seemed excessive.

api/server.go
```diff
@@ -98,14 +99,25 @@ func (s *Server) walletAddressHandler(w http.ResponseWriter, req *http.Request,
...
+	begin, end := beginEnd(req.URL.Query(), len(addresses))
+	writeJSON(w, WalletAddresses(addresses[begin:end]))
 }
 
 func (s *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	transactions := s.w.Transactions()
-	offset, limit := offsetLimit(req.URL.Query(), len(transactions))
-	writeJSON(w, transactions[offset:limit])
+	q := req.URL.Query()
+	since := q.Get("since")
```
lukechampine, line 108: for query string parameters we generally use `req.FormValue("since")`

api/server.go
```diff
@@ -98,14 +99,25 @@ func (s *Server) walletAddressHandler(w http.ResponseWriter, req *http.Request,
 
 func (s *Server) walletAddressesHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
 	addresses := s.w.Addresses()
-	offset, limit := offsetLimit(req.URL.Query(), len(addresses))
-	writeJSON(w, WalletAddresses(addresses[offset:limit]))
+	begin, end := beginEnd(req.URL.Query(), len(addresses))
+	writeJSON(w, WalletAddresses(addresses[begin:end]))
```
lukechampine, line 103: I'd probably inline `beginEnd`, so that you can have better error messages:

```go
addresses := s.w.Addresses()
start, end := 0, len(addresses)
if v := req.FormValue("start"); v != "" {
    i, err := strconv.Atoi(v)
    if err != nil || i < 0 || i > len(addresses) {
        http.Error(w, "invalid start value", http.StatusBadRequest)
        return
    }
    start = i
}
if v := req.FormValue("end"); v != "" {
    i, err := strconv.Atoi(v)
    if err != nil || i < 0 || i < start {
        http.Error(w, "invalid end value", http.StatusBadRequest)
        return
    } else if i > len(addresses)
        i = len(addresses)
    }
    end = i
}
writeJSON(w, WalletAddresses(addresses[begin:end]))
```

api/server.go
```diff
@@ -98,14 +99,25 @@ func (s *Server) walletAddressHandler(w http.ResponseWriter, req *http.Request,
...
 }
 
 func (s *Server) walletTransactionsHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
-	transactions := s.w.Transactions()
-	offset, limit := offsetLimit(req.URL.Query(), len(transactions))
-	writeJSON(w, transactions[offset:limit])
+	q := req.URL.Query()
+	since := q.Get("since")
+	sinceParsed, err := strconv.ParseInt(since, 10, 64)
```
lukechampine, line 109: RFC3339 is probably a better choice than Unix timestamps:
```go
var since time.Time
if v := req.FormValue("since"); v != "" {
    t, err := time.Parse(time.RFC3339, req.FormValue("since")
    if err != nil {
        http.Error(w, "invalid since value: " + err.Error(), http.StatusBadRequest)
        return
    }
}

File: api/server.go [hunk 4]
```diff
@@ -98,14 +99,25 @@ func (s *Server) walletAddressHandler(w http.ResponseWriter, req *http.Request,
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	max := q.Get("max")
+	maxParsed, err := strconv.Atoi(max)
+	if err != nil && max != "" {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	writeJSON(w, s.w.Transactions(time.Unix(sinceParsed, 0), maxParsed))
```
lukechampine, line 120: `maxParsed` should have a default value of `-1`, so that `Transactions` returns everything.

internal/walletutil/store.go
```diff
@@ -65,10 +66,17 @@ func (s *EphemeralStore) SpendableSiacoinElements() []types.SiacoinElement {
...
 // oldest-to-newest.
-func (s *EphemeralStore) Transactions() []wallet.Transaction {
+func (s *EphemeralStore) Transactions(since time.Time, max int) (txns []wallet.Transaction) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	return append([]wallet.Transaction(nil), s.txns...)
+	for _, txn := range s.txns {
+		if len(txns) == max {
+			return
+		} else if txn.Timestamp.Unix() >= since.Unix() {
```
lukechampine, line 75: `txn.Timestamp.After(since)`

(technically, this is equivalent to `>`, not `>=`, but *also* technically, that might be the more correct interpretation of the word "since"...? Perhaps we should change "since" to "start" if we want it to unambiguously refer to a closed interval.)

lukechampine (commented):
some minor feedback. Don't forget to update the `api.Client` methods as well.

chris124567:
I figured if we are going to use standard POST forms we might as well just use JSON since that's what all the other endpoints are using.  So I just defined some types for the wallet addresses/transactions requests and then we just unmarshal those on the server side.  This feels cleaner to me than using `strconv`.

lukechampine:
sry for the confusion -- `req.FormValue`, even though it seems like it would only work for `POST` requests, will actually read values from the query string if no form data is present. IMO these are best represented as `GET` requests. That seems to be standard practice for other APIs.

PR #59, opened 2021-12-10
v2: Add syncer store interface and ensure syncer has a healthy number of peers

chris124567:
This implements a minimal interface used by the syncer to store peer addresses.  Currently only an in memory store is supported but a store that writes to the disk will be added shortly.  It also adds a maintainHealthyPeerSet method on Syncer which runs as a goroutine that retrieves peers from the store and connects to them if the peer count is less than 8.

### DIFF
--- a/cmd/siad/node.go
+++ b/cmd/siad/node.go
@@ -11,6 +11,7 @@ import (
 	"go.sia.tech/core/consensus"
 	"go.sia.tech/siad/v2/internal/chainutil"
 	"go.sia.tech/siad/v2/internal/cpuminer"
+	"go.sia.tech/siad/v2/internal/p2putil"
 	"go.sia.tech/siad/v2/internal/walletutil"
 	"go.sia.tech/siad/v2/p2p"
 	"go.sia.tech/siad/v2/txpool"
@@ -80,6 +81,15 @@ func newNode(addr, dir string, c consensus.Checkpoint) (*node, error) {
 	}
 	seed := wallet.NewSeed()
 
+	syncerDir := filepath.Join(dir, "p2p")
+	if err := os.MkdirAll(p2pDir, 0700); err != nil {
+		return nil, err
+	}
+	syncerStore, err := p2putil.NewJSONStore(syncerDir)
+	if err != nil {
+		return nil, err
+	}
+
 	cm := chain.NewManager(chainStore, tip.Context)
 	tp := txpool.New(tip.Context)
 	cm.AddSubscriber(tp, cm.Tip())
@@ -91,7 +101,7 @@ func newNode(addr, dir string, c consensus.Checkpoint) (*node, error) {
 	m := cpuminer.New(tip.Context, w.NextAddress(), tp)
 	cm.AddSubscriber(m, cm.Tip())
 
-	s, err := p2p.NewSyncer(addr, genesisBlock.ID(), cm, tp)
+	s, err := p2p.NewSyncer(addr, genesisBlock.ID(), cm, tp, syncerStore)
 	if err != nil {
 		return nil, err
 	}
--- a/p2p/syncer.go
+++ b/p2p/syncer.go
@@ -234,11 +234,19 @@ func rpcDeadline(id rpc.Specifier) time.Duration {
 	}
 }
 
+// A SyncerStore stores peer addresses. Implementations are assumed to be thread
+// safe.
+type SyncerStore interface {
+	AddPeer(addr string) error
+	RandomPeer() (string, error)
+}
+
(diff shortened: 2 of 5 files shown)


lukechampine:
I think you forgot to commit `store.go` 😅

chris124567:
There we go :)

p2p/store.go
```diff
@@ -0,0 +1,43 @@
...
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	// map iteration order in go is random
+	for peer := range e.peers {
+		return peer, nil
+	}
+	return "", errors.New("no peers in list")
+}
+
+func NewEphemeralStore(peers []string) *EphemeralStore {
```
lukechampine, line 35: I would drop the argument here. You can initialize the store by calling `AddPeer`.

p2p/syncer.go
```diff
@@ -239,11 +239,13 @@ type Syncer struct {
...
 
-	cond  sync.Cond
-	mu    sync.Mutex
-	peers map[gateway.Header]*gateway.Session
-	err   error
+	cond     sync.Cond
+	peerCond sync.Cond
+	mu       sync.Mutex
+	peers    map[gateway.Header]*gateway.Session
+	err      error
```
lukechampine, line 248: let's stick with one cond for now. Wakeups are going to be fairly uncommon, and spurious wakeups are cheap -- the goroutine just checks a few variables and goes back to sleep.

p2p/syncer.go
```diff
@@ -536,6 +538,34 @@ func (s *Syncer) syncToPeer(peer gateway.Header) {
...
+		if err != nil {
+			continue
+		}
+
+		exists := false
+		for _, connected := range s.Peers() {
+			if peer == connected.NetAddress {
+				exists = true
+			}
+		}
```
lukechampine, line 560: for this, you'll want to create a `seen := make(map[string]bool)` at the top of `maintainHealthyPeerSet`. This will prevent you from repeatedly attempting to connect to unreachable peers.

p2p/store.go
```diff
@@ -0,0 +1,43 @@
+package p2p
```
lukechampine, line 1: store implementations (especially ephemeral ones) should live in an `internal` package, in this case `internal/p2putil`. We'll probably make them publicly visible eventually, but for now I want to keep the exposed API minimal.

lukechampine (changes requested):
looks alright. Pls also add a `JSONStore`, using the same approach as in https://github.com/SiaFoundation/siad/blob/v2-devel/internal/walletutil/store.go

chris124567:
I think this should do it.

cmd/siad/node.go
```diff
@@ -80,6 +81,15 @@ func newNode(addr, dir string, c consensus.Checkpoint) (*node, error) {
 	}
 	seed := wallet.NewSeed()
 
+	p2pDir := filepath.Join(dir, "p2p")
+	if err := os.MkdirAll(p2pDir, 0700); err != nil {
+		return nil, err
+	}
+	p2pStore, err := p2putil.NewJSONStore(p2pDir)
```
lukechampine, line 88: I'd call these `syncerDir` and `syncerStore`

p2p/syncer_test.go
```diff
@@ -104,7 +105,11 @@ func newTestNode(tb testing.TB, genesisID types.BlockID, c consensus.Checkpoint)
 	cm.AddSubscriber(ws, cm.Tip())
 	m := cpuminer.New(c.Context, w.NextAddress(), tp)
 	cm.AddSubscriber(m, cm.Tip())
-	s, err := NewSyncer(":0", genesisID, cm, tp, NewEphemeralStore(nil))
+	js, err := p2putil.NewJSONStore(".")
```
lukechampine, line 108: can use an ephemeral store here

p2p/syncer.go
```diff
@@ -536,6 +542,30 @@ func (s *Syncer) syncToPeer(peer gateway.Header) {
...
+func (s *Syncer) maintainHealthyPeerSet() {
+	seen := make(map[string]struct{})
+	for {
+		s.cond.L.Lock()
+		for len(s.peers) >= 8 {
+			// cond is signalled upon disconnecting from a peer
+			s.cond.Wait()
+		}
+		peer, err := s.store.RandomPeer()
+		s.cond.L.Unlock()
```
lukechampine, line 555: not necessary to call `RandomPeer` while holding the lock; the store has its own internal mutex. Probably the `SyncerStore` docstring should note that implementations are assumed to be thread-safe. (This is also true of `wallet.Store`.)

PR #60, opened 2021-12-14
v2: Add peer discovery mechanism and RPC methods

chris124567:
This adds a GetPeers method to the syncer RPC interface.  Clients requests the amount of peers they want and the server sends back the addresses of that many peers from their peer list.  Then, in maintainHealthyPeerSet, if we are already connected to the the random peer from the store, we call getPeers on a random one of our peers.

### DIFF
--- a/p2p/syncer.go
+++ b/p2p/syncer.go
@@ -27,11 +27,15 @@ var errInternalError = errors.New("could not complete request due to internal er
 
 var (
 	rpcGetHeaders    = rpc.NewSpecifier("GetHeaders")
+	rpcGetPeers      = rpc.NewSpecifier("GetPeers")
 	rpcGetBlocks     = rpc.NewSpecifier("GetBlocks")
 	rpcGetCheckpoint = rpc.NewSpecifier("GetCheckpoint")
 	rpcRelayBlock    = rpc.NewSpecifier("RelayBlock")
 	rpcRelayTxn      = rpc.NewSpecifier("RelayTxn")
 )
+
+// Maximum number of peers a getPeers request can return.
+const maxGetPeers = 100
 
 type msgGetHeaders struct {
 	History []types.ChainIndex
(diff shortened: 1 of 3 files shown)


p2p/syncer.go
```diff
@@ -442,6 +494,22 @@ func (s *Syncer) handleRPC(peer gateway.Header, msg rpc.Object) (resp rpc.Object
...
 		return &msgHeaders{Headers: headers}, nil
 
+	case *msgGetPeers:
+		count := msg.Count
+		if count > uint64(len(s.peers)) {
+			count = uint64(len(s.peers))
+		}
+
+		peers := make([]string, count)
+		for peer := range s.peers {
```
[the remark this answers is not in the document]
lukechampine: I guess using `RandomPeer()` means there's a chance of sending duplicate peers. Actually, it's worse than that: if you only have one peer, you'll send that single peer, duplicated 100 times!

It's not pretty, but probably the best way to handle this is to add a `RandomPeers(n int)` method to `SyncerStore`.

p2p/syncer.go
```diff
@@ -560,6 +630,27 @@ func (s *Syncer) maintainHealthyPeerSet() {
...
 			time.Sleep(time.Second)
+
+			// if we're already connected to the random peer from the store,
+			// try to request additional peers
+			s.mu.Lock()
+			var randomPeer gateway.Header
+			for peer := range s.peers {
+				randomPeer = peer
+				break
+			}
```
lukechampine, line 641: this can be:
```go
var randomPeer gateway.Header
for randomPeer = range s.peers {
    break
}
```

p2p/syncer.go
```diff
@@ -560,6 +630,27 @@ func (s *Syncer) maintainHealthyPeerSet() {
 		} else {
 			// sleep on failure to avoid spinning unproductively
 			time.Sleep(time.Second)
+
+			// if we're already connected to the random peer from the store,
+			// try to request additional peers
```
lukechampine, line 635: We should assume that the majority of `Connect` calls will fail, meaning we'll end up on this branch quite frequently. So calling `getPeers` each time is going to be overkill. We should probably keep track of which peers we've called `getPeers` on, and ensure that we only call it once per peer.

chris124567:
Alright.  I think we should just get started directly on the PeerManager next.

PR #63, opened 2021-12-15
v2: add CI and linting

n8maninger:
+ Copies the siafoundation/core golangci and test files
+ Fixes lint errors


lukechampine:
Race should be fixed by https://github.com/SiaFoundation/siad/commit/5232b092581b3890bc899186bf80a1cb44513db5; try rebasing

n8maninger:
> Race should be fixed by [5232b09](https://github.com/SiaFoundation/siad/commit/5232b092581b3890bc899186bf80a1cb44513db5); try rebasing

almost

lukechampine:
gah, ok, I tested https://github.com/SiaFoundation/siad/commit/d09032df7efb173c598ac5fdf9218f628beeacd7 locally and it didn't trigger the race detector, so hopefully that works.

lukechampine:
I added some locking and extended the timeout on the failing test. Try again? 🙃 

More broadly, it looks like I should really revisit the `ChainManager` API. It was a decent design for the single-threaded message-queue p2p code of `sunyata`, but I don't think it's a good fit for a highly-concurrent RPC-based architecture.

n8maninger:
Using `go test -failfast -count=10 ./...`,  `TestNetwork` fails almost every time on localhost. Getting all 6 jobs to pass on the same run seems unlikely.

lukechampine:
dang. Well, I certainly don't want to go back to the dark ages of rerunning tests over and over because they fail nondeterministically. I guess we'll have to actually get to the bottom of why `TestNetwork` is failing.

n8maninger:
If we're going to be redoing parts of the p2p, do you think we should go ahead and merge this? It would help keep new inconsistent tests out while we sort out the p2p code.

lukechampine:
I'm gonna try to fix the NDF first; no sense merging this if it's gonna fail on every PR

lukechampine:
phew, ok, I think I really tracked it down this time. Since we're mining at low difficulty, there was a chance that two different chains could have exactly the same amount of total work. Nodes will only reorg to a competing chain if it has *strictly more* work than their own, so the two test nodes would both reject each other's chains and remain desynced.

In real life, this isn't a problem because A) the work values are much larger, and so the probability of an exact match is much lower, and B) if there *is* an exact match, it will resolve itself once another block is mined. AFAIK most blockchain implementations, including Bitcoin, don't even bother handling the case where work is exactly equal.

I also fixed a race condition in the txpool. I've run `TestNetwork` 100 times locally without any failures so... I *would* say we're good now, but I'm afraid to anger the gods any further.

PR #66, opened 2021-12-23
v2: Add peer manager to handle peer misbehavior

chris124567:
This adds a new PeerManager interface designed to be used to handle misbehaving peers.  In the p2putil.DefaultPeerManager, peers start out with 0 failed connections and a banscore of 0.  If we have problems communicating with them over RPC or the data they give us is bad, their banscore is increased.   Once the banscore exceeds maxBanscore, they are blacklisted, disconnected from, and cannot be connected to again.  If we have more than maxFailedConnects failed connections with them, we also blacklist them and stop trying to connect to them.  It also keeps track of the last interaction time with the peer as advised in https://github.com/SiaFoundation/siad/pull/60#pullrequestreview-832007297 but I am not exactly sure what to use this for.  Also, right now this code works on a per NetAddress (IP + port) basis but I think it should be purely on a IP basis to increase the cost of spam/harming the network.

### DIFF
--- a/internal/p2putil/store.go
+++ b/internal/p2putil/store.go
@@ -6,62 +6,95 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+
+	"go.sia.tech/siad/v2/p2p"
 )
 
-// EphemeralStore implements p2p.SyncerStore in memory.
+// EphemeralStore implements p2p.PeerStore in memory.
 type EphemeralStore struct {
+	peers map[string]p2p.PeerInfo
 	mu    sync.Mutex
-	peers map[string]struct{}
 }
 
-// AddPeer implements p2p.SyncerStore.
+// AddPeer implements p2p.PeerStore.
 func (s *EphemeralStore) AddPeer(addr string) error {
 	s.mu.Lock()
-	s.peers[addr] = struct{}{}
-	s.mu.Unlock()
+	defer s.mu.Unlock()
+	p := p2p.PeerInfo{
+		NetAddress: addr,
+	}
+	if _, ok := s.peers[addr]; !ok {
+		s.peers[addr] = p
+	}
 	return nil
 }
 
-// RandomPeer implements p2p.SyncerStore.
-func (s *EphemeralStore) RandomPeer() (string, error) {
+// Info implements p2p.PeerStore.
+func (s *EphemeralStore) Info(addr string) p2p.PeerInfo {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	for peer := range s.peers {
-		return peer, nil
+	p, ok := s.peers[addr]
+	if !ok {
+		p = p2p.PeerInfo{
+			NetAddress: addr,
+		}
 	}
-	return "", errors.New("no peers in list")
+	return p
 }
 
-// RandomPeers implements p2p.SyncerStore.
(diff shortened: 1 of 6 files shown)


p2p/syncer.go
```diff
@@ -809,17 +808,16 @@ func (s *Syncer) Close() error {
 }
 
 // NewSyncer creates a new syncer listening on the given address.
-func NewSyncer(addr string, genesisID types.BlockID, cm *chain.Manager, tp *txpool.Pool, store SyncerStore, pm PeerManager) (*Syncer, error) {
+func NewSyncer(addr string, genesisID types.BlockID, cm *chain.Manager, tp *txpool.Pool, store PeerStore, npm NewPeerManager) (*Syncer, error) {
```
lukechampine, line 811: `npm`!! 😱

lukechampine: I think the interfaces are a bit confused here. When you have a constructor that takes a function and then call that function on another argument to the constructor...that's usually a bad sign.

Let's step back and ask: does `PeerManager` need to be an interface at all? I think not; its behavior is always the same. The only thing we want to be dynamic is the `PeerStore`. So let's move `p2putil.DefaultPeerManager` into `p2p`, and just call it `PeerManager`. Then, remove the `store` parameter from `NewSyncer`; it doesn't need it. Now, our initialization flow looks like:

```go
ps := p2putil.NewEphemeralStore()
pm := p2p.NewPeerManager(ps)
s := p2p.NewSyncer(addr, genesisID, cm, tp, pm)
```

chris124567:
Alright now we just have one peer manager and no interface.  I can't reproduce that test failure even with `-race`.  If I had to guess it's that the time.Sleep calls in the test aren't long enough

lukechampine:
I ended up tweaked the `PeerStore` interface a bit. One thing that occurred to me is that sometimes we want to group peers by their hostname (e.g. if we want to blacklist a spammer more effectively), but we don't *always* want to do this (e.g. when testing locally with multiple peers all on `localhost`). So we may want the peer manager to track hostname separately from individual peers. Not sure yet; in any case, it's not a high priority.

Another future direction would be to report a lot more peer metadata in the API. For reference, [look how much info `bitcoind` gives you](https://developer.bitcoin.org/reference/rpc/getpeerinfo.html). We could extend `p2p.PeerInfo` with these fields and call `Info` on each address returned from `(Syncer).Peers`.

internal/p2putil/store.go
```diff
@@ -6,65 +6,98 @@ import (
...
 	s.EphemeralStore.AddPeer(addr)
 	return s.save()
 }
 
+// UpdatePeer implements p2p.PeerStore.
+func (s *JSONStore) UpdatePeer(addr string, fn func(*p2p.PeerInfo)) error {
+	if err := s.EphemeralStore.UpdatePeer(addr, fn); err != nil {
+		return err
+	}
+	return s.save()
```
lukechampine, line 98: Saving after every update is fine for now, but when the `Syncer` is busy this will generate tons of unnecessary I/O. Saving after every `AddPeer` call is similarly inefficient, since we might add 100 peers at a time. Ideally, we would only save in certain cases, e.g. when blacklisting, or after a certain number of updates, or after a certain period of quiescence.

PR #69, opened 2022-01-06
Add /contract endpoints

chris124567:
This adds a `muse` like API to the new siad.  This PR depends on code implemented in https://github.com/SiaFoundation/core/pull/24 so we should wait to merge until that is finalized.

### DIFF
--- a/api/renterd/api.go
+++ b/api/renterd/api.go
@@ -1,6 +1,9 @@
 package renterd
 
-import "go.sia.tech/core/types"
+import (
+	"go.sia.tech/core/net/rhp"
+	"go.sia.tech/core/types"
+)
 
 // WalletBalanceResponse is the response to /wallet/balance. It contains the confirmed
 // Siacoin and Siafund balance of the wallet.
@@ -43,4 +46,48 @@ type SyncerPeerResponse struct {
 // A SyncerConnectRequest requests that the syncer connect to a peer.
 type SyncerConnectRequest struct {
 	NetAddress string `json:"netAddress"`
+}
+
+// A ContractsScanRequest contains the information of the host we are dialing
+// for contract related requests.
+type ContractsScanRequest struct {
+	NetAddress string          `json:"netAddress"`
+	HostKey    types.PublicKey `json:"hostKey"`
+}
+
+// A ContractsFormRequest requests that the host create a contract.
+type ContractsFormRequest struct {
+	RenterKey    types.PrivateKey `json:"renterKey"`
+	NetAddress   string           `json:"netAddress"`
+	HostKey      types.PublicKey  `json:"hostKey"`
+	HostFunds    types.Currency   `json:"hostFunds"`
+	RenterFunds  types.Currency   `json:"renterFunds"`
+	EndHeight    uint64           `json:"endHeight"`
+	HostSettings rhp.HostSettings `json:"hostSettings"`
+}
+
+// A ContractsFormResponse is the response to /contract/form. It contains the
+// formed contract and the parent transactions.
+type ContractsFormResponse struct {
+	Contract rhp.Contract      `json:"contract"`
+	Parent   types.Transaction `json:"parent"`
+}
+
+// A ContractsRenewResponse is the response to /contract/renew. It contains the
+// formed contract and the parent transactions.
+type ContractsRenewResponse struct {
+	Contract rhp.Contract      `json:"contract"`
+	Parent   types.Transaction `json:"parent"`
+}
+
+// A ContractsRenewRequest requests that the host renew a contract.
+type ContractsRenewRequest struct {
+	RenterKey      types.PrivateKey           `json:"renterKey"`
+	NetAddress     string                     `json:"netAddress"`
+	HostKey        types.PublicKey            `json:"hostKey"`
+	RenterFunds    types.Currency             `json:"renterFunds"`
+	HostCollateral types.Currency             `json:"hostCollateral"`
+	Extension      uint64                     `json:"extension"`
+	Contract       types.FileContractRevision `json:"contract"`
+	HostSettings   rhp.HostSettings           `json:"hostSettings"`
 }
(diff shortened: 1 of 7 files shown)


api/api.go
```diff
@@ -46,3 +47,37 @@ type SyncerPeer struct {
 type SyncerConnectRequest struct {
 	NetAddress string `json:"netAddress"`
 }
+
+// A ContractBaseRequest contains the information of the host we are dialing
+// for contract related requests.
+type ContractBaseRequest struct {
+	NetAddress string          `json:"netAddress"`
+	TheirKey   types.PublicKey `json:"theirKey"`
+}
```
lukechampine, line 56: I'd just call this `ScanRequest`, and inline the fields in `FormRequest` and `RenewRequest`

api/api.go
```diff
@@ -46,3 +47,37 @@ type SyncerPeer struct {
...
+type ContractFormResponse struct {
+	Contract renter.Contract     `json:"contract"`
+	Parents  []types.Transaction `json:"parents"`
+}
+
+// A ContractRenewRequest requests that the host renew a contract.
+type ContractRenewRequest struct {
+	ContractBaseRequest
+	RenterKey             types.PrivateKey            `json:"renterKey"`
+	Contract              *types.FileContractRevision `json:"contract"`
```
lukechampine, line 79: why use a pointer here?

api/client.go
```diff
@@ -104,6 +105,24 @@ func (c *Client) SyncerConnect(addr string) (err error) {
 	return
 }
 
+// ContractScan scans a host and retrieves its settings.
+func (c *Client) ContractScan(netAddress string, theirKey types.PublicKey) (resp rhp.HostSettings, err error) {
```
lukechampine, line 109: `theirKey` should be `hostKey` (here and elsewhere)

api/client.go
```diff
@@ -104,6 +105,24 @@ func (c *Client) SyncerConnect(addr string) (err error) {
...
+	err = c.post("/contract/scan", ContractBaseRequest{netAddress, theirKey}, &resp)
+	return
+}
+
+// ContractForm forms a contract.
+func (c *Client) ContractForm(netAddress string, theirKey types.PublicKey, renterKey types.PrivateKey, hostFunds, renterFunds types.Currency, endHeight uint64) (resp ContractFormResponse, err error) {
+	err = c.post("/contract/form", ContractFormRequest{ContractBaseRequest{netAddress, theirKey}, renterKey, hostFunds, renterFunds, endHeight}, &resp)
```
lukechampine, line 116: this is a pretty large struct; best to write it out one-line-per-field

chris124567:
Rebased and added support for host sets.  I just left the import to go.sia.tech/core/renter even though it is no longer present because a similar thing (after it was decided that specific RPC handling code like that shouldn't go in `core`) is being worked on in siad and I will replace those references once it is ported.

chris124567:
Rebased to support the recent renterd PR.  Moved all the API methods to the renterd api module.  The level of code duplication between renterd/siad makes me uncomfortable but I guess there may not be a better way.

api/renterd/api.go
```diff
@@ -49,6 +57,10 @@ type server struct {
 	s  Syncer
 	cm ChainManager
 	tp TransactionPool
+
+	mu        sync.Mutex
+	contracts []rhp.Contract
+	hostSets  map[string][]types.PublicKey
```
lukechampine, line 63: for now, we can just return the contract in the API response and not save anything server-side. We can skip the host set stuff too; I'm not sure it's the best design

api/renterd/api.go
```diff
@@ -188,6 +200,144 @@ func (s *server) syncerConnectHandler(w http.ResponseWriter, req *http.Request,
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	cm := (s.cm).(*chain.Manager)
+	session, err := renter.NewSession(cfr.NetAddress, cfr.HostKey, s.w, s.tp, cm)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
lukechampine, line 236: we'll want to add a `defer session.Close()` here

api/renterd/api.go
```diff
@@ -210,5 +360,13 @@ func NewServer(cm ChainManager, s Syncer, w Wallet, tp TransactionPool) http.Han
 	mux.GET("/api/syncer/peers", srv.syncerPeersHandler)
 	mux.POST("/api/syncer/connect", srv.syncerConnectHandler)
 
+	mux.POST("/contract/scan", srv.contractScanHandler)
+	mux.POST("/contract/form", srv.contractFormHandler)
+	mux.POST("/contract/renew", srv.contractRenewHandler)
```
lukechampine, line 365: needs an `/api` prefix

api/renterd/requests.go
```diff
@@ -44,3 +47,49 @@ type SyncerPeerResponse struct {
...
+	EndHeight   uint64           `json:"endHeight"`
+}
+
+// A ContractFormResponse is the response to /contract/form and
+// /contract/renew.  It contains the formed contract and the parent
+// transactions.
+type ContractFormResponse struct {
+	Contract rhp.Contract        `json:"contract"`
+	Parents  []types.Transaction `json:"parents"`
+}
```
lukechampine, line 74: My gut says that renew should have a separate `Response` type, even if they currently contain the same fields.

api/renterd/api.go
chris124567: As far as I can tell Session does not have a Close method

chris124567:
Tried to resolve the merge conflicts with the Github website and it did this... one second.

lukechampine:
It's actually fine (and preferable) to define `ChainManager` in both places; the real `*chain.Manager` will satisfy both interfaces.

chris124567:
These errors are strange.  I don't get them on my system (go1.16.10 linux/amd64) and it doesn't really make sense to me why 

```
assignment mismatch: 1 variable but s.cm.TipContext returns 2 values
```

is the error for this line (which assigns two variables):

```
    vc, err := s.cm.TipContext()
```
when all definitions of ChainManager are:

```
	// A ChainManager manages blockchain state.
	ChainManager interface {
		TipContext() (consensus.ValidationContext, error)
	}
```

chris124567:
Finally :) Should be good to go.

PR #71, opened 2022-01-11
ci: add nightly v2 testing

n8maninger:
Runs the test suite nightly with -count=1000 to try and detect NDFs. The workflow must be on the master branch for scheduling, but the `v2-devel` branch will be checked out on run.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,28 @@
+name: Nightly v2
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  test:
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v2
+        with:
+          ref: 'v2-devel' # checkout v2-devel branch, scheduled actions only run on the default branch
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: test
+        uses: n8maninger/action-golang-test@v1
+        with:
+          args: "-race;-count=1000;-timeout=30m"
```
lukechampine, line 28: `100` has been sufficient for me, especially if we're enabling `race`. I would also add `-failfast`

PR #72, opened 2022-01-11
Use RPCReadMaxLen for reading subscription requests

ChrisSchinnerl:
Some users have seen the error `2022/01/11 06:04:47.824071 errors.go:150: [failed to handle request; failed to read subscription request; length 13641 exceeds maxLen of 4096]` in their logs because clients like to subscribe to many registry entries at once and 4096 will only let them subscribe to around 40 at the same time.

That's why I suggest increasing that limit because the user is paying for the used ram anyway.

### DIFF
--- a/modules/host/rpcsubscribe.go
+++ b/modules/host/rpcsubscribe.go
@@ -44,10 +44,9 @@ type (
 	subscriptionInfoID types.Specifier
 )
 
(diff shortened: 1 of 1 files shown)


lukechampine:
Seems reasonable enough to me

PR #73, opened 2022-01-12
v2: Add support for Siafunds in wallet

chris124567:
This PR adds support for Siafunds in the wallet interface.  It also updates the `siac` commands accordingly.  There's lots of code duplication here and I'm not sure if there exists an elegant way around it.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -72,6 +72,7 @@ type Store interface {
 	AddAddress(addr types.Address, index uint64) error
 	AddressIndex(addr types.Address) (uint64, bool)
 	SpendableSiacoinElements() []types.SiacoinElement
+	SpendableSiafundElements() []types.SiafundElement
 	Transactions(since time.Time, max int) []Transaction
 }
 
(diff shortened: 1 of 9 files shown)


lukechampine:
IDs are unique across SC/SF, i.e. there will never be an SC input with the same ID as an SF input. That means you should be able to have a single `SignTransaction` function that works for either SC or SF (or both).

wallet/wallet.go
```diff
@@ -180,8 +189,74 @@ func (w *HotWallet) FundTransaction(txn *types.Transaction, amount types.Currenc
 	return toSign, discard, nil
 }
 
-// SignableInputs returns the inputs of txn that the wallet can sign.
-func (w *HotWallet) SignableInputs(txn types.Transaction) []types.ElementID {
+// FundTransactionSiafund adds Siafund inputs worth at least amount to txn. It
+// returns the IDs of the added inputs, as well as a "discard" function that,
+// when called releases the inputs for use in other transactions.
+func (w *HotWallet) FundTransactionSiafund(txn *types.Transaction, amount types.Currency, pool []types.Transaction) ([]types.ElementID, func(), error) {
```
lukechampine, line 195: `amount` should be a `uint64`

chris124567:
I rebased it to the latest and made it so it could get balances only.  Also added code so ProcessChainRevertUpdate would apply correctly to SF elements in the store.

chris124567:
Rebased again to fix merge conflicts - should be fine to merge now.

lukechampine:
Thanks for keeping this updated. My current thinking is that siafunds should be handled by a separate, non-hot wallet, and only supported by `siad` (not `renterd` or others). I have an in-progress version of this (based on your code) locally. When it's ready I'll push to this branch.

chris124567:
No longer relevant after wallet refactor.

PR #74, opened 2022-01-14
Add Pool.AcceptTransactionSet and HotWallet.Address

n8maninger:
+ Adds `Pool.AcceptTransactionSet` to validate and add a transaction set to the transaction pool
+ Adds `HotWallet.Address` to get the address of the last address added to the wallet

Off-topic, but should the store interface be changed to work on spend policies rather than only indices?

### DIFF
--- a/txpool/txpool.go
+++ b/txpool/txpool.go
@@ -2,6 +2,7 @@ package txpool
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 
 	"go.sia.tech/core/chain"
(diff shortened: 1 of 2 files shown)


lukechampine:
A generic store probably would, yeah, but the current store is meant for use with seed-based, non-multisig wallets. We should certainly provide wallets that support hardware wallets and multisig, but `HotWallet` covers 99% of usecases. I guess we should at least rename `Store` to `SeedStore`, though.

PR #79, opened 2022-02-01
Check Siafund and file contract revision ephemeral outputs 

n8maninger:


lukechampine:
see https://github.com/SiaFoundation/core/pull/43

PR #80, opened 2022-02-08
WIP: Add explored

chris124567:
This implements a basic new binary in `cmd` that will eventually keep track of the addresses/outputs in the chain.  Right now it only has a /api/consensus endpoint that returns some information from the chain manager's validation context.


api/explored/api.go
```diff
@@ -0,0 +1,119 @@
...
+		return
+	}
+
+	if err := s.s.Connect(scr.NetAddress); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+}
+
+func (s *server) consensusHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
```
lukechampine, line 87: maybe call this `consensusTip`, with route `/api/consensus/tip`

api/explored/api.go
```diff
@@ -0,0 +1,119 @@
...
+	ctx := s.cm.TipContext()
```
lukechampine, line 88: canonical name for a validation context is `vc`

(`ctx` usually means a `context.Context`)

api/explored/api.go
```diff
@@ -0,0 +1,119 @@
...
+		ctx.Difficulty,
+		ctx.OakWork,
+		ctx.OakTime,
+		ctx.GenesisTimestamp,
+		ctx.SiafundPool,
+		ctx.FoundationAddress,
+	})
+}
+
+// NewServer returns an HTTP handler that serves the siad API.
```
lukechampine, line 101: siad -> explored (here and in a few other places too)

api/explored/requests.go
```diff
@@ -0,0 +1,39 @@
+package explored
```
lukechampine, line 1: my preferred file layout for api stuff is:

- request/response types in `api.go`
- server implementation in `server.go`
- client implementation in `client.go`

cmd/explored/main.go
```diff
@@ -0,0 +1,152 @@
...
+	bootstrap := flag.String("bootstrap", "", "peer address or explorer URL to bootstrap from")
+	flag.Parse()
+
+	log.Println("explored v0.1.0")
+	if flag.Arg(0) == "version" {
+		return
+	}
+
+	initCheckpoint := genesis
+	if *checkpoint != "" {
```
lukechampine, line 63: we can probably remove this functionality from the explorer; explorers should track the history of the entire chain, so it doesn't make much sense for them to bootstrap from a recent checkpoint

api/explored/requests.go
lukechampine: yeah, should change all of them

PR #82, opened 2022-02-16
RHP use renewal resolution 

n8maninger:
~~Updates the FileContract signatures from SiaFoundation/core#47 and changes the renter side of the protocol to send the signatures earlier for SiaFoundation/core#55. Will need changes from SiaFoundation/core#54 and possibly SiaFoundation/core#51 before merging.~~

Updates the renter side of the protocol to use the renewal resolution from SiaFoundation/core#51

Lint and tests will fail until `go.sia.tech/core` is updated in go.mod


n8maninger:
Relies on SiaFoundation/core#56

lukechampine:
the renewal calculations have always felt too abstruse to me; I'm gonna punt for now and just have the caller supply the payouts as arguments

renter/rpc.go
```diff
@@ -206,3 +206,271 @@ func (s *Session) ScanSettings() (rhp.HostSettings, error) {
...
+		hostInputs[i].Signatures = hostSigs.SiacoinInputSignatures[i]
+	}
+
+	// the transaction should now be valid
+	if err := vc.ValidateTransaction(renewalTxn); err != nil {
+		return rhp.Contract{}, types.Transaction{}, fmt.Errorf("renewal transaction is invalid: %w", err)
+	}
+
+	return rhp.Contract{
+		ID:       renewalTxn.FileContractID(0),
```
lukechampine, line 473: I opted to just call `FileContractID` here instead of constructing the ID manually. We could also add something like `FileContractRenewalID` but that feels awkward. I think reusing `FileContractID` is Fine, Actually, as long as we document it in the docstring.

PR #83, opened 2022-02-17
Update FileContract fields

n8maninger:
~~Adds the initial cost to the renter for data already stored in the contract back to the Renew RPC. For a renewal that extends the duration of a contract, the renter must pay additional storage and the host must risk additional collateral.~~

Updates FileContract usage in the RHP to match SiaFoundation/core#57

### DIFF
--- a/renter/rpc.go
+++ b/renter/rpc.go
@@ -219,17 +219,17 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 		return rhp.Contract{}, types.Transaction{}, errors.New("end height must be greater than start height")
 	}
 
-	hostPayout := hostCollateral.Add(settings.ContractFee)
+	hostPayout := settings.ContractFee.Add(hostCollateral)
 	fc := types.FileContract{
-		Filesize:           0,
-		WindowStart:        endHeight,
-		WindowEnd:          endHeight + settings.WindowSize,
-		ValidRenterOutput:  types.SiacoinOutput{Value: renterFunds, Address: renterAddr},
-		ValidHostOutput:    types.SiacoinOutput{Value: hostPayout, Address: hostAddr},
-		MissedRenterOutput: types.SiacoinOutput{Value: renterFunds, Address: renterAddr},
-		MissedHostOutput:   types.SiacoinOutput{Value: hostPayout, Address: hostAddr},
-		RenterPublicKey:    renterKey.PublicKey(),
-		HostPublicKey:      s.hostKey,
+		Filesize:        0,
+		WindowStart:     endHeight,
+		WindowEnd:       endHeight + settings.WindowSize,
+		RenterOutput:    types.SiacoinOutput{Value: renterFunds, Address: renterAddr},
+		HostOutput:      types.SiacoinOutput{Value: hostPayout, Address: hostAddr},
+		MissedHostValue: hostPayout,
+		TotalCollateral: hostCollateral,
+		RenterPublicKey: renterKey.PublicKey(),
+		HostPublicKey:   s.hostKey,
 	}
 
 	if err := rhp.ValidateContractFormation(fc, startHeight, settings); err != nil {
@@ -244,7 +244,7 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 	}
 	txn.MinerFee = s.tpool.RecommendedFee().Mul64(vc.TransactionWeight(txn))
 	// fund the formation transaction with the renter allowance, contract fee,
-	// siafund tax, and miner fee.
+	// siafund tax, and miner fee
 	totalCost := renterFunds.Add(settings.ContractFee).Add(vc.FileContractTax(fc)).Add(txn.MinerFee)
 	toSign, cleanup, err := s.wallet.FundTransaction(&txn, totalCost, nil)
 	if err != nil {
@@ -276,14 +276,14 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 		return rhp.Contract{}, types.Transaction{}, errors.New("host signature is invalid")
 	}
 
-	// add the host's additions to the transaction.
+	// add the host's additions to the transaction
 	renterInputs := txn.SiacoinInputs
 	txn.SiacoinInputs = append(txn.SiacoinInputs, resp.Inputs...)
 	hostInputs := txn.SiacoinInputs[len(renterInputs):]
 	txn.SiacoinOutputs = append(txn.SiacoinOutputs, resp.Outputs...)
 	txn.FileContracts[0].HostSignature = resp.ContractSignature
 
-	// sign the transaction and send the signatures to the host.
+	// sign the transaction and send the signatures to the host
 	if err := s.wallet.SignTransaction(vc, &txn, toSign); err != nil {
 		return rhp.Contract{}, types.Transaction{}, fmt.Errorf("failed to sign transaction: %w", err)
 	}
@@ -297,7 +297,7 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 		return rhp.Contract{}, types.Transaction{}, fmt.Errorf("failed to write renter signatures: %w", err)
 	}
 
-	// read the host's signatures and add them to the transaction.
+	// read the host's signatures and add them to the transaction
 	var hostSigs rhp.RPCContractSignatures
 	if err := rpc.ReadResponse(stream, &hostSigs); err != nil {
 		return rhp.Contract{}, types.Transaction{}, fmt.Errorf("failed to read host signatures: %w", err)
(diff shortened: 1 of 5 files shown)


renter/rpc.go
```diff
@@ -362,20 +391,24 @@ func (s *Session) RenewContract(renterKey types.PrivateKey, contract types.FileC
...
+	// The renter is responsible for it's additional funding, the host's
+	// contract fee, the base storage cost, siafund tax, and miner fee.
+	totalCost := additionalRenterFunds.Add(settings.ContractFee).Add(baseStorageCost).Add(vc.FileContractTax(renewalTxn.FileContractResolutions[0].Renewal.InitialRevision)).Add(renewalTxn.MinerFee)
 	renterRollover := finalRevision.ValidRenterOutput.Value
+	var toSign []types.ElementID
 	if renterRollover.Cmp(totalCost) < 0 {
-		// there's isn't enough value left in the old contract to pay for the
-		// new one; isn't enough to pay for the new contract; add more inputs
+		// there is not enough value left in the old contract to pay for the new
+		// one; add more inputs.
```
lukechampine, line 401: what do you have against contractions

renter/rpc.go
```diff
@@ -308,16 +308,21 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
...
 }
 
-// RenewContract renews an existing contract, modifying its funding and/or
-// duration. Funds remaining in the old contract will be rolled over into the
-// new contract.
-func (s *Session) RenewContract(renterKey types.PrivateKey, contract types.FileContractRevision, renterFunds, validHostPayout, missedHostPayout types.Currency, endHeight uint64) (rhp.Contract, types.Transaction, error) {
+// RenewContract renews an existing contract, adding additional funds,
+// collateral and/or duration. Funds remaining in the old contract will be
+// rolled over into the new contract.
+func (s *Session) RenewContract(renterKey types.PrivateKey, contract types.FileContractRevision, additionalRenterFunds, additionalHostCollateral types.Currency, endHeight uint64) (rhp.Contract, types.Transaction, error) {
```
lukechampine, line 325: I'd prefer to leave this more generic:
```go
func (s *Session) RenewContract(renterKey types.PrivateKey, contract types.FileContractRevision, renterFunds, validHostCollateral, missedHostCollateral types.Currency, endHeight uint64)
```
I'm just not super confident that the renewal math is correct for all cases. Better to punt those calculations up the stack; then we can implement wrappers like `ExtendContract` (increase endheight)  and `RefillContract` (keep same endheight) with more straightforward math, and we won't have to keep tweaking this API.

n8maninger: There are only three possible scenarios: the renter is extending the length of the contract, adding funds, or both.

Correct me if I'm wrong, but I don't believe we can use that method signature because it doesn't capture the portion of the renter's funds that should be allocated to the valid host output to pay for the storage extension. Since the host has to validate the renewal pays them enough, we have to get the math correct on that end regardless.

lukechampine: In principle, you could have a renewal with a shorter duration and less funds! The only thing that distinguishes a renewal from a new contract is that the host is already storing some data (and thus, should be compensated for storing that data, and should be liable for losing that data). If the host greatly lowers their prices after the initial contract formation, then you could end up paying less for a renewal even if it extends the duration. So I don't like calling these parameters `additional`, because the amount of funds/collateral isn't necessarily going to increase.

I guess the host parameters should be `validHostPayout` and `missedHostPayout` (since they aren't *purely* collateral), but in any case we know that only three currency values are needed (since the renter payout is the same for both valid and missed).

n8maninger: Renewals for shorter durations don’t affect the existing storage cost though. The renter has already payed for the duration at the time of upload as part of the existing contract. I’m not sure it’s worth it to change the channel so we can provide an accurate refund 🤔

Currently, the only part the renter *can* change is how much additional funding they want to add to the contract and it’s expiration. Then we need to determine how much additional storage time they’re using and pay the host for that. So renter funds is really two variables, renter storage extension cost and renter unused funds

lukechampine: hmm. I guess it simplifies things to just ban renewals with a lower end height. In that case, we should use an `extension` parameter rather than `endHeight`.

Maybe things would be clearer if we defined `FileContract` like so:
```go
type FileContract struct {
    // other fields omitted

    RenterOutput   types.SiacoinOutput
    HostOutput     types.SiacoinOutput

    HostCollateralAvailable types.Currency
    HostCollateralRisked    types.Currency
}
```
We initialize the contract with `RenterOutput = renterFunds`, `HostOutput = settings.ContractFee`, `HostCollateralAvailable = hostCollateral`, and `HostCollateralRisked = 0`. The renter pays for `RenterOutput + HostOutput`, and the host pays for `HostCollateralAvailable`. To revise the contract, we move value from `RenterOutput` to `HostOutput`; if we're uploading, we also move value from `HostCollateralAvailable` to `HostCollateralRisked`. If the contract resolves valid, then we disburse `RenterOutput` and `HostOutput + HostCollateralAvailable + HostCollateralRisked`; but if it resolves invalid, then we only disburse `HostOutput + HostCollateralAvailable`.

So: what does renewal look like with this design?

The renter can freely calculate new values for `RenterOutput` and `HostCollateralAvailable`, based on how much it wants to spend on the contract and how much collateral it wants the host to make available. The initial `HostOutput` should be `settings.ContractFee`, as before, and `HostCollateralRisked` should be recalculated as `filesize * settings.Collateral`. The renter funds the transaction with `RenterOutput + HostOutput - renterRollover`, and the host funds `HostCollateralAvailable + HostCollateralRisked - hostRollover`.

There's just one annoyance, though. When we renew a contract, we're also finalizing the old contract. In the old contract, the renter was paying the host to store data for X blocks -- but now, we're renewing after only Y blocks, so we're overpaying the host!

I see two ways of dealing with this. First, we can correct the overpayment in the final revision; to do this, move `filesize * storage cost * remaining duration` from the host back to the renter. Second, we can require the host to rollover *at least this amount* into the new contract. Both have tradeoffs; open to suggestions here.

n8maninger: > If the contract resolves valid, then we disburse RenterOutput and HostOutput + HostCollateralAvailable + HostCollateralRisked; but if it resolves invalid, then we only disburse HostOutput + HostCollateralAvailable.

Shouldn't `RenterOutput` always be disbursed since it's the *unspent* renter funds? Would we need a different value to track the spent and burnt funds?

Hmm, does this take into account the renter's burnt allowance?

> When we renew a contract, we're also finalizing the old contract. In the old contract, the renter was paying the host to store data for X blocks -- but now, we're renewing after only Y blocks, so we're overpaying the host!

If the renter only pays for the extension, it's not overpaying the host. The renter only pays the difference between the original and the new end. The host isn't risking as much collateral though since the risk would be the extension duration, not the extension + remaining.

lukechampine: yeah, I omitted it for brevity, but the renter gets their output regardless. Not sure what you mean about the burn field?

By "overpayment," I'm referring to how the finalization releases outputs immediately. If the renter paid the host to store for 10 blocks, but renews after 5 blocks, then the finalization output will still pay the host as though they stored for 10 blocks.

renter/rpc.go
```diff
@@ -308,46 +308,57 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 		hostInputs[i].Signatures = hostSigs.SiacoinInputSignatures[i]
 	}
 
+	// the transaction should now be valid
+	if err := vc.ValidateTransaction(txn); err != nil {
```
lukechampine, line 312: Hmm. This is tricky. If the renter and host aren't on the same block, then their input proofs will be out-of-sync, and the transaction will be invalid. Worse, we can't simply lean on the txpool's "flexibility" in updating proofs, because the txpool will call `UpdateElementProof` on *all* of the elements. So to make this work, we need the renter and host to agree on a validation context to use.

This feels like it will be pretty ugly to deal with properly. On the bright side, the Utreexo proofs aren't actually *creating* a new problem -- they're revealing a problem that was always there. Any consensus mismatch between the renter and host creates the potential for "races" like this that can result in invalid transactions.

lukechampine: So we'll need to add some form of synchronization to the RHP. Maybe the best API for this would be for the renter and host to send their `vc.Index` alongside their inputs. Then, we can add a generic proof-updating function to `ChainManager`:

```go
func (cm *ChainManager) UpdateElementProof(e *types.StateElement, cur, dst types.ChainIndex) error
```

renter/rpc.go
```diff
@@ -276,14 +276,14 @@ func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostColl
 		return rhp.Contract{}, types.Transaction{}, errors.New("host signature is invalid")
 	}
 
-	// add the host's additions to the transaction.
+	// add the host's additions to the transaction
```
lukechampine, line 279: my (admittedly eclectic) commenting style is:

- treat "simple" comments (up to 1 sentence) as sentence fragments: no initial capital, no period
- multi-sentence comments should use proper capitalization and punctuation
- liberally abuse semicolons, em-dashes, and parentheticals in order to prevent a single sentence from becoming multiple

renter/rpc.go
lukechampine: I went ahead and implemented this in https://github.com/SiaFoundation/core/commit/40ae1d57010c5ac814d72ea8c988161953918c97

IMO we should just merge this PR as-is so that #69 can proceed, and tackle the sync issue in a follow up.

n8maninger: Any value we send, we have to deal with a small race between the call to `cm.TipContext` and `w.FundTransaction`... The easiest fix would be to send the renter's `ChainIndex` in the initial `ContractRequest` and make the host slightly flexible, like the txpool. Created #85 for discussion and LGTM.

PR #86, opened 2022-02-28
v2: Flesh out explored

chris124567:
This adds a subscriber to the chain manager that gets information about new blocks and stores that information in a database.  Since the database we are going to use is still up in the air, a new Explorer needs a current validation context and an explorer.Store.  The store must implement methods like:

```
	AddSiacoinElement(sce types.SiacoinElement) error
	AddSiafundElement(sce types.SiafundElement) error
...
	BlockHeight() (uint64, error)
	SetBlockHeight(height uint64) error
```
etc.  Any database can be used by just making a wrapper around it and implementing those methods.  Importantly, we keep information on each siacoin/siafund/contract element and a summary of each block called BlockFacts.  Since consensus.ApplyUpdate's don't contain transactions, we are limited in some ways compared to the old siad (i.e., it is no longer possible to find all the transactions containing a certain output).


api/explored/server.go
```diff
@@ -114,5 +196,11 @@ func NewServer(cm ChainManager, s Syncer, tp TransactionPool) http.Handler {
 
 	mux.GET("/api/consensus/tip", srv.consensusTipHandler)
 
+	mux.POST("/api/explorer/element/siacoin", srv.explorerElementSiacoinHandler)
+	mux.POST("/api/explorer/element/siafund", srv.explorerElementSiafundHandler)
+	mux.POST("/api/explorer/element/contract", srv.explorerElementContractHandler)
+	mux.POST("/api/explorer/block/facts", srv.explorerBlockFactsHandler)
```
lukechampine, line 202: should be `GET`, with the ID or height as part of the url

explorer/info.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"time"
+
+	"go.sia.tech/core/types"
+)
+
+// BlockFacts returns a bunch of statistics about the consensus set as they
+// were at a specific block.
+type BlockFacts struct {
```
lukechampine, line 11: It'll probably take a few iterations before we settle on the right data to store. I've asked Hakkane (of SiaStats) what kind of API he'd like to see in the new explorer, but in the meantime we can at least be sure about storing the full block and its full validation context.

explorer/info.go
```diff
@@ -0,0 +1,82 @@
...
+func (e *Explorer) LatestBlockFacts() (BlockFacts, bool) {
+	height, err := e.db.BlockHeight()
+	if err != nil {
+		return BlockFacts{}, false
+	}
+	return e.BlockFacts(height)
+}
+
+// BlockFacts returns facts about the black at the the specified height.
+func (e *Explorer) BlockFacts(height uint64) (BlockFacts, bool) {
```
lukechampine, line 49: should take a `types.ChainIndex`, since it should be possible to request the facts for any block (not just blocks on the best chain)

explorer/info.go
```diff
@@ -0,0 +1,82 @@
...
+	facts, err := e.db.BlockFacts(height)
+	if err != nil {
+		return BlockFacts{}, false
+	}
+	return facts, true
+}
+
+// SiacoinElement returns the siacoin element associated with the specified ID.
+func (e *Explorer) SiacoinElement(id types.ElementID) (types.SiacoinElement, bool) {
```
lukechampine, line 58: these should all return the original error, not a `bool` (idk why the old implementation does that)

explorer/info.go
```diff
@@ -0,0 +1,82 @@
...
+	RevisedFileContractsCount  uint64
+	ResolvedFileContractsCount uint64
+
+	// Facts about file contracts.
+	ActiveContractCost  types.Currency
+	ActiveContractCount uint64
+	ActiveContractSize  uint64
+	TotalContractCost   types.Currency
+	TotalContractSize   uint64
+	TotalRevisionVolume uint64
```
lukechampine, line 36: I would rename this struct `ChainStats`, and remove the fields taken from `ValidationContext` -- they can be handled separately

explorer/update.go
```diff
@@ -0,0 +1,189 @@
...
+	mu sync.Mutex
+	vc consensus.ValidationContext
+}
+
+// ProcessChainApplyUpdate implements chain.Subscriber.
+func (e *Explorer) ProcessChainApplyUpdate(cau *chain.ApplyUpdate, _ bool) error {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	oldFacts, err := e.db.BlockFacts(e.vc.Index.Height)
```
lukechampine, line 41: consider adding a `tipStats  ChainStats` field to `Explorer` instead of retrieving them from the db each time

explorer/update.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	for _, elem := range cru.NewFileContracts {
+		if err := e.db.RemoveElement(elem.ID); err != nil {
+			return err
+		}
+	}
+
+	if err := e.db.RemoveBlockFacts(e.vc.Index.Height); err != nil {
+		return err
+	}
```
lukechampine, line 173: we'll want to keep the stats for all blocks

explorer/update.go
```diff
@@ -0,0 +1,189 @@
...
+			return err
+		}
+	}
+	for _, elem := range cru.NewSiafundElements {
+		if err := e.db.RemoveElement(elem.ID); err != nil {
+			return err
+		}
+	}
+	for _, elem := range cru.RevisedFileContracts {
+		if err := e.db.RemoveElement(elem.ID); err != nil {
```
lukechampine, line 161: the contract still exists even if a revision is reverted. It should be reset to its prior state (`elem.Parent`)

explorer/update.go
```diff
@@ -0,0 +1,189 @@
...
+	AddFileContractElement(fce types.FileContractElement) error
+	RemoveElement(id types.ElementID) error
+	SiacoinElement(id types.ElementID) (types.SiacoinElement, error)
+	SiafundElement(id types.ElementID) (types.SiafundElement, error)
+	FileContractElement(id types.ElementID) (types.FileContractElement, error)
+	BlockFacts(height uint64) (BlockFacts, error)
+	AddBlockFacts(height uint64, facts BlockFacts) error
+	RemoveBlockFacts(height uint64) error
+	BlockHeight() (uint64, error)
+	SetBlockHeight(height uint64) error
```
lukechampine, line 25: thinking about it some more, this API could cause performance headaches. If every `Add` or `Remove` call is fsync'ing the disk, and we call those functions dozens of times per update, then performance will be terrible.

I can think of two solutions:
- Add a `Commit` method, called when the `mayCommit` argument of `ProcessChainApplyUpdate` is set
- Consolidate the methods into just `AddChainUpdate` and `RevertChainUpdate`, moving much of the logic from `ProcessChainApplyUpdate`/`ProcessChainRevertUpdate` into the interface implementation.

The database we choose may end up influencing this decision, so let's postpone it for now.

cmd/explored/node.go
```diff
@@ -48,6 +50,8 @@ func newNode(addr, dir string, c consensus.Checkpoint) (*node, error) {
 	cm := chain.NewManager(chainStore, tip.Context)
 	tp := txpool.New(tip.Context)
 	cm.AddSubscriber(tp, cm.Tip())
+	e := explorer.New(tip.Context, nil)
+	cm.AddSubscriber(e, cm.Tip())
```
lukechampine, line 54: this error needs to be checked, and it should use the *explorer's* tip, not the chain manager's tip. `AddSubscriber` tries to "migrate" the subscriber from its current tip to `cm.Tip()`, by calling `ProcessChainApplyUpdate` and `ProcessChainRevertUpdate` as needed. So if you pass `cm.Tip()`, no work is necessary (and thus no error is possible) -- but in most cases, we want to pass the subscriber's tip.

lukechampine (commented):
neat. btw, you can access the transactions via `cau.Block`/`cru.Block`

chris124567:
Alright I think that's all of it.  I'm unsure if I handled revisions properly in the case where the block is being reverted (see https://github.com/SiaFoundation/siad/pull/86/commits/e0a7feeee9f61d413ca7699514770553ae90f23a#diff-1de3df26962ac7217113b9b0b671f85c395eb5e03b8ff66a35327dcae42d7aaeL164).  I remove all the revised contracts, then add back the parents.

n8maninger:
We should track address balance and transactions. Not being able to query the balance of any address was a sticking point for the last exchange that needed help integrating. Maybe a couple of `[POST] /batch` endpoints for addresses and transactions?

api/explored/server.go
```diff
@@ -114,5 +196,11 @@ func NewServer(cm ChainManager, s Syncer, tp TransactionPool) http.Handler {
 
 	mux.GET("/api/consensus/tip", srv.consensusTipHandler)
 
+	mux.GET("/api/explorer/element/siacoin", srv.explorerElementSiacoinHandler)
```
n8maninger, line 199: I generally prefer `/siacoin/:id` over query params for endpoints like this. Do you have a preference @lukechampine?

lukechampine: yeah, any sort of stable ID (element ID, txid, block ID) should be part of the URL

chris124567:
What kind of batch endpoints specifically do you want? Just added the other suggestions.

n8maninger:
Address balance, unspent outputs, and transactions is the primary use-case where batching requests is useful. Maybe:
+ `[POST] /batch/addresses/balance` - returns the SF and SC balance of the addresses
+ `[POST] /batch/addresses/siacoins` - returns the unspent SCE of the addresses
+ `[POST] /batch/addresses/siafunds` - returns the unspent SFE of the addresses
+ `[POST] /batch/addresses/transactions` - returns the last *n* transactions of the addresses

chris124567:
Added those batch endpoints.  I guess the next step should be to implement the `Store` with sqlite or some other database.

chris124567:
I added a basic store using sqlite (the modernc.org CGO-free implementation).  I think next we should make it so that all actions (like AddChainStats, AddTransaction etc) are added to a transaction that is then committed at the end of ProcessChainApplyUpdate/ProcessChainRevertUpdate to avoid excessive writes.  Also need to test the store more.

go.sum
```diff
@@ -1,26 +1,193 @@
...
+modernc.org/opt v0.1.1/go.mod h1:WdSiB5evDcignE70guQKxYUl14mgWtbClRi5wmkkTX0=
+modernc.org/sqlite v1.14.8 h1:2OOqfZAyU4x4qusilvHoRXXqsAgaZobi1o+mjQ5MUpw=
+modernc.org/sqlite v1.14.8/go.mod h1:TFmXjym+/jR31fxc2B5eHnKMuJJGY7i1L/T5A0jzVww=
+modernc.org/strutil v1.1.1 h1:xv+J1BXY3Opl2ALrBwyfEikFAj8pmqcpnfmuwUwcozs=
+modernc.org/strutil v1.1.1/go.mod h1:DE+MQQ/hjKBZS2zNInV5hhcipt5rLPWkmpbGeW5mmdw=
+modernc.org/tcl v1.11.0/go.mod h1:zsTUpbQ+NxQEjOjCUlImDLPv1sG8Ww0qp66ZvyOxCgw=
+modernc.org/token v1.0.0 h1:a0jaWiNMDhDUtqOj09wvjWWAqd3q7WpBulmL9H2egsk=
+modernc.org/token v1.0.0/go.mod h1:UGzOrNV1mAFSEB63lOFHIpNRUVMvYTc6yu1SMY/XTDM=
+modernc.org/z v1.3.0/go.mod h1:+mvgLH814oDjtATDdT3rs84JnUIpkvAF5B8AVkNlE2g=
+modernc.org/z v1.3.1/go.mod h1:0RBFPpdFNiKpjTza1WYaB4+6ySjS6dLBoo09OQZ4E3w=
```
lukechampine, line 193: good lord lol. Does `go mod tidy` remove most of these?

chris124567: Just ran go mod tidy and pushed again.  Wonder why so many versions of the same package are still included.

internal/explorerutil/store.go
```diff
@@ -0,0 +1,321 @@
...
+);
+
+CREATE TABLE unspentElements (
+    id VARCHAR(128),
+    type VARCHAR(128),
+	address VARCHAR(128)
+);
+
+CREATE TABLE transactions (
+    id VARCHAR(128),
```
n8maninger, line 44: `id` should be a `PRIMARY KEY`

internal/explorerutil/store.go
```diff
@@ -0,0 +1,321 @@
...
+    data BLOB
+);
+
+CREATE TABLE addressTransactions (
+    id VARCHAR(128),
```
n8maninger, line 49: If the lib supports foreign keys, we should use them https://www.sqlite.org/foreignkeys.html

internal/explorerutil/store.go
```diff
@@ -0,0 +1,321 @@
...
+func NewStore(dir string) (*SqliteStore, error) {
+	db, err := sql.Open("sqlite", path.Join(dir, "store.db"))
+	if err != nil {
+		return nil, err
+	}
+	if _, err := db.Exec(`
+CREATE TABLE elements (
+    id VARCHAR(128),
+    type VARCHAR(128),
+    data BLOB
```
n8maninger, line 29: Probably best to include a `NOT NULL` constraint on `DATA`

go.sum
n8maninger: embarcadero's CI scripts added `windows/arm64` as a build target, but it looks like this lib doesn't support it yet https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures https://gitlab.com/cznic/sqlite/-/issues/81. We'll need to keep that in mind when we add `explored`'s build scripts.

n8maninger:
embarcadero's CI scripts added windows/arm64 as a build target, but it looks like this lib doesn't support it yet https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures https://gitlab.com/cznic/sqlite/-/issues/81. We'll need to keep that in mind when we add explored's build scripts.

chris124567:
Fixed lots of things and added a test that covers `explorer.SiacoinBalance`, `explorer.UnspentSiacoinElements`, `explorer.SiacoinElement`, and `explorer.Transactions`.  I Would have made more comprehensive tests but the SpendPolicy field in inputs is causing problems with `explorer.ChainStats`/`explorer.Transaction` because it cannot be JSON unmarshaled (`cannot unmarshal array into Go struct field SiacoinInput.SiacoinInputs.SpendPolicy of type types.SpendPolicy`).  Since it is an interface type we cannot just do something like:

```
func (p *SpendPolicy) UnmarshalJSON(b []byte) err {
    d := types.NewDecoder(bytes.NewReader(b))
    d.ReadPolicy(p)
    return nil
}
```

We could create some kind of wrapper type like

```
type X struct {
    SpendPolicy
}
```

but that feels sort of hacky.

chris124567:
Sounds good - just solved that problem.  I'll add some other tests mostly focusing on making sure the ChainStats statistics are accurate.

api/explored/client.go
```diff
@@ -40,6 +44,127 @@ func (c *Client) ConsensusTip() (resp ConsensusTipResponse, err error) {
...
+}
+
+// ExplorerBatchSiafunds returns the unspent siafund elements of the addresses.
+func (c *Client) ExplorerBatchSiafunds(addresses []types.Address) (resp [][]types.ElementID, err error) {
+	err = c.c.Post("/api/explorer/batch/addresses/siafunds", addresses, &resp)
+	return
+}
+
+// ExplorerBatchTransactions returns the last n transactions of the addresses.
+func (c *Client) ExplorerBatchTransactions(addresses []ExplorerTransactionsRequest) (resp [][]types.ElementID, err error) {
```
n8maninger, line 163: Ideally, this would return the transaction data, not just ids.

api/explored/client.go
```diff
@@ -40,6 +44,127 @@ func (c *Client) ConsensusTip() (resp ConsensusTipResponse, err error) {
 	return
 }
 
+// ExplorerSiacoinElement gets the Siacoin element with the given ID.
+func (c *Client) ExplorerSiacoinElement(id types.ElementID) (resp types.SiacoinElement, err error) {
+	data, err := json.Marshal(id)
```
lukechampine, line 49: should use `id.String()` for these

api/explored/client.go
```diff
@@ -40,6 +44,127 @@ func (c *Client) ConsensusTip() (resp ConsensusTipResponse, err error) {
...
+}
+
+// ExplorerChainStatsLatest gets stats about the latest block.
+func (c *Client) ExplorerChainStatsLatest() (resp explorer.ChainStats, err error) {
+	err = c.c.Get("/api/explorer/chain/stats/latest", &resp)
+	return
+}
+
+// ExplorerSearch gets information about a given element.
+func (c *Client) ExplorerSearch(id types.ElementID) (resp ExplorerSearchResponse, err error) {
```
lukechampine, line 94: `Search` is a bit too generic; we can call this `Element`

api/explored/client.go
```diff
@@ -40,6 +44,127 @@ func (c *Client) ConsensusTip() (resp ConsensusTipResponse, err error) {
...
+	data, err := json.Marshal(id)
+	if err != nil {
+		return
+	}
+	err = c.c.Get(fmt.Sprintf("/api/explorer/search/%s", string(data)), &resp)
+	return
+}
+
+// ExplorerBalance gets the siacoin and siafund balance of an address.
+func (c *Client) ExplorerBalance(address types.Address) (resp ExplorerWalletBalanceResponse, err error) {
```
lukechampine, line 104: `ExplorerAddressBalance`

(a good rule of thumb is that the method name should match the API route)

api/explored/server.go
```diff
@@ -97,12 +116,295 @@ func (s *server) consensusTipHandler(w http.ResponseWriter, req *http.Request, _
 	})
 }
 
+func (s *server) explorerElementSiacoinHandler(w http.ResponseWriter, req *http.Request, p httprouter.Params) {
+	var id types.ElementID
+	if err := json.Unmarshal([]byte(p.ByName("id")), &id); err != nil {
```
lukechampine, line 121: `id.UnmarshalText` is probably the right choice here

explorer/info.go
```diff
@@ -0,0 +1,158 @@
...
+	ValidationContext consensus.ValidationContext
+
+	// Transaction type counts.
+	SpentSiacoinsCount         uint64
+	SpentSiafundsCount         uint64
+	NewSiacoinsCount           uint64
+	NewSiafundsCount           uint64
+	NewFileContractsCount      uint64
+	RevisedFileContractsCount  uint64
+	ResolvedFileContractsCount uint64
```
lukechampine, line 21: It's unclear to me whether these counts are very useful in practice. Let's remove them for now.

explorer/info.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/types"
+)
+
+// ChainStats contains a bunch of statistics about the consensus set as they
+// were at a specific block.
+type ChainStats struct {
+	Block             types.Block
+	ValidationContext consensus.ValidationContext
```
lukechampine, line 12: Probably don't want to be encoding an entire ValidationContext as JSON. We can add a `/chain/context/` route for querying them.

explorer/update.go
```diff
@@ -0,0 +1,250 @@
+package explorer
```
lukechampine, line 1: this file should be named `explorer.go`

explorer/update.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		stats.SpentSiafundsCount++
+	}
+	for _, elem := range cau.ResolvedFileContracts {
+		if err := e.db.RemoveElement(elem.ID); err != nil {
+			return err
+		}
+		stats.ResolvedFileContractsCount++
+		stats.ActiveContractCount--
+		payout := elem.FileContract.MissedHostValue.Add(elem.FileContract.TotalCollateral)
```
lukechampine, line 114: should be computed as `RenterOutput.Value.Add(HostOutput.Value)`

explorer/update.go
```diff
@@ -0,0 +1,250 @@
...
+		stats.RevisedFileContractsCount++
+		stats.TotalContractSize += elem.FileContract.Filesize
+		stats.TotalRevisionVolume += elem.FileContract.Filesize
+	}
+	for _, elem := range cau.NewFileContracts {
+		if err := e.db.AddFileContractElement(elem); err != nil {
+			return err
+		}
+		stats.NewFileContractsCount++
+		payout := elem.FileContract.MissedHostValue.Add(elem.FileContract.TotalCollateral)
```
lukechampine, line 150: ditto

explorer/update.go
```diff
@@ -0,0 +1,250 @@
...
+	for _, elem := range cau.SpentSiafunds {
+		if err := e.db.RemoveElement(elem.ID); err != nil {
+			return err
+		}
+		if err := e.db.RemoveUnspentSiafundElement(elem.Address, elem.ID); err != nil {
+			return err
+		}
+		stats.SpentSiafundsCount++
+	}
+	for _, elem := range cau.ResolvedFileContracts {
```
lukechampine, line 108: we'll have to account for all the different types of resolution here; see the docstring of `FileContractResolution`

lukechampine (commented):
we should document the DB schema somewhere; maybe in a block comment in `store.go`?

explorer/update.go
chris124567: I got everything except this.  Do you mean accounting for renewals?

api/explored/client.go
```diff
@@ -90,18 +79,18 @@ func (c *Client) ExplorerChainStatsLatest() (resp explorer.ChainStats, err error
...
-// ExplorerSearch gets information about a given element.
-func (c *Client) ExplorerSearch(id types.ElementID) (resp ExplorerSearchResponse, err error) {
+// ExplorerElementSearch gets information about a given element.
+func (c *Client) ExplorerElementSearch(id types.ElementID) (resp ExplorerSearchResponse, err error) {
 	data, err := json.Marshal(id)
 	if err != nil {
 		return
 	}
-	err = c.c.Get(fmt.Sprintf("/api/explorer/search/%s", string(data)), &resp)
+	err = c.c.Get(fmt.Sprintf("/api/explorer/element/search/%s", string(data)), &resp)
```
lukechampine, line 88: same here, use `id.String()`

api/explored/client.go
```diff
@@ -165,6 +154,16 @@ func (c *Client) ExplorerBatchTransactions(addresses []ExplorerTransactionsReque
 	return
 }
 
+// ExplorerChainContext returns the validation context at a given chain index.
+func (c *Client) ExplorerChainContext(index types.ChainIndex) (resp consensus.ValidationContext, err error) {
+	data, err := json.Marshal(index)
+	if err != nil {
+		return
+	}
+	err = c.c.Get(fmt.Sprintf("/api/explorer/chain/context/%s", string(data)), &resp)
```
lukechampine, line 163: and here

chris124567:
Rebased to the latest v2-devel.

lukechampine:
Moved to https://github.com/SiaFoundation/explorer

PR #89, opened 2022-03-08
Ignore immature outputs in TestWallet balance

n8maninger:
It's inconvenient for testing to have immature outputs counted in the balance since they can't be used.

### DIFF
--- a/internal/walletutil/store.go
+++ b/internal/walletutil/store.go
@@ -37,7 +37,9 @@ func (s *EphemeralStore) Balance() (sc types.Currency, sf uint64) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	for _, sce := range s.scElems {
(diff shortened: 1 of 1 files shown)


lukechampine:
Yeah, this is tricky because we also don't want to e.g. give miners the impression that they aren't receiving block rewards. Similarly, it could be confusing for a host to see that they submitted a valid storage proof but their balance didn't change.

Maybe the best way to handle this is to explicitly track block rewards and file contracts, like [`walrus` does](https://lukechampine.com/docs/walrus/#list-block-rewards). Then the respective UIs can just show those objects directly.

lukechampine:
It's already possible to compute `TimelockedBalance` via `UnspentSiacoinOutputs`; just a bit awkward. We can revisit this when we start the UI design in earnest.

PR #97, opened 2022-03-15
Flesh out core renter RPCs

lukechampine:
Adds the `Lock`, `Read`, and `Write` (well, `Append`) RPCs to `renter.Session`, and wires the latter up to the API. Also renames the `/contract` routes to `/rhp`, and adds `Client` methods for all `/rhp` endpoints.

This should provide all the functionality we need to build a minimal renter UI.

### DIFF
--- a/api/renterd/server.go
+++ b/api/renterd/server.go
@@ -2,12 +2,15 @@ package renterd
 
 import (
 	"encoding/json"
+	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"time"
 
 	"github.com/julienschmidt/httprouter"
 	"go.sia.tech/core/consensus"
+	"go.sia.tech/core/net/rhp"
 	"go.sia.tech/core/types"
 	"go.sia.tech/siad/v2/api"
 	"go.sia.tech/siad/v2/internal/walletutil"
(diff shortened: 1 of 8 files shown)


renter/rpc.go
```diff
@@ -183,34 +217,151 @@ func (s *Session) RegisterSettings(payment PaymentMethod) (settings rhp.HostSett
...
+	return nil
+}
+
+// Read calls the Read RPC, writing the requested sections of sector data to w.
+// Merkle proofs are always requested.
+//
+// Note that sector data is streamed to w before it has been validated. Callers
+// MUST check the returned error, and discard any data written to w if the error
+// is non-nil. Failure to do so may allow an attacker to inject malicious data.
+func (s *Session) Read(ctx context.Context, w io.Writer, sections []rhp.RPCReadRequestSection) (err error) {
```
n8maninger, line 279: Does a separate `Read` or a session-wide `Lock` still make sense?

lukechampine: perhaps not, but I'm resisting the urge to make any significant changes to the RHP for now. We'll probably overhaul a lot of stuff as a result of the upcoming summit.

renter/rpc.go
```diff
@@ -183,34 +217,151 @@ func (s *Session) RegisterSettings(payment PaymentMethod) (settings rhp.HostSett
...
+	}
+
+	return nil
+}
+
 // FormContract negotiates a new contract with the host using the specified
 // funds and duration.
-func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostCollateral types.Currency, endHeight uint64, settings rhp.HostSettings) (rhp.Contract, types.Transaction, error) {
-	renterAddr := s.wallet.Address()
+func (s *Session) FormContract(renterKey types.PrivateKey, renterFunds, hostCollateral types.Currency, endHeight uint64, settings rhp.HostSettings, w Wallet, tp TransactionPool) (rhp.Contract, types.Transaction, error) {
```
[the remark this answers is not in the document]
lukechampine: I think I wrote it that way in an early version of `us`, but it bothered me that you couldn't be sure exactly which block your contract would end on. That's why I generally prefer to specify absolutely rather than relatively when possible. (`extension` is relative, but it's relative to a fixed value, so arguably it's fixed too.)

api/renterd/server.go
```diff
@@ -177,12 +180,95 @@ func (s *server) contractsRenewHandler(w http.ResponseWriter, req *http.Request,
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	defer session.Close()
+	if err := session.Lock(req.Context(), rrr.ContractID, rrr.RenterKey); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	var refundAccountID types.PublicKey // TODO: ???
```
lukechampine, line 208: @n8maninger what pubkey should we be using for these?

n8maninger: It's an EA account; any key pair will work. I've been using `renterKey.Public()` for simplicity while testing; it's probably OK not to expose anything more advanced in renterd.

api/renterd/server.go
```diff
@@ -177,12 +180,93 @@ func (s *server) contractsRenewHandler(w http.ResponseWriter, req *http.Request,
 
 	// TODO: compare request settings to current host settings
 
-	contract, parent, err := session.RenewContract(crr.RenterKey, crr.Contract, crr.RenterFunds, crr.HostCollateral, crr.Extension)
+	contract, parent, err := session.RenewContract(rrr.RenterKey, rrr.Contract, rrr.RenterFunds, rrr.HostCollateral, rrr.Extension, s.w, s.tp)
```
n8maninger, line 183: `RenewContract` also requires a call to `RegisterSettings`

PR #103, opened 2022-04-01
ci: add v1.18

n8maninger:


.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        go-version: [ '1.16', '1.17' ]
+        go-version: [ '1.16', '1.17', '1.18' ]
```
lukechampine, line 21: To keep CI times fast, I'd prefer to restrict our matrix to two versions.

n8maninger: Do you want to stick with 1.16 and 1.17 or current and current - 1?

PR #105, opened 2022-04-26
Add domain and ip range blocking to HostDB's filtering system

chris124567:
Adds domain and IP address filtering capabilities to `SetFilterMode` in the HostDB.  Currently has some bugs that need to be ironed out.

### DIFF
--- a/cmd/siac/hostdbcmd.go
+++ b/cmd/siac/hostdbcmd.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"strings"
 	"text/tabwriter"
 
 	"github.com/spf13/cobra"
@@ -37,11 +38,13 @@ var (
 	}
 
 	hostdbSetFiltermodeCmd = &cobra.Command{
-		Use:   "setfiltermode [filtermode] [host] [host] [host]...",
+		Use:   "setfiltermode [filtermode] [host,...] [netaddress,...]",
 		Short: "Set the filtermode.",
-		Long: `Set the hostdb filtermode and specify hosts.
+		Long: `Set the hostdb filtermode and specify comma separated hosts and net addresses.
         [filtermode] can be whitelist, blacklist, or disable.
-        [host] is the host public key.`,
+        [host] is the host public key.
+        [netaddress] is the host domain, IP address, or IP address range.
+        `,
 		Run: hostdbsetfiltermodecmd,
 	}
 
@@ -327,6 +330,10 @@ func hostdbfiltermodecmd() {
 	for _, host := range hdfmg.Hosts {
 		fmt.Println("    ", host)
 	}
+	fmt.Println("  Net Addresses:")
+	for _, netAddress := range hdfmg.NetAddresses {
+		fmt.Println("    ", netAddress)
+	}
 	fmt.Println()
 }
 
@@ -337,6 +344,7 @@ func hostdbsetfiltermodecmd(cmd *cobra.Command, args []string) {
 	var filterModeStr string
 	var host types.SiaPublicKey
 	var hosts []types.SiaPublicKey
+	var netAddresses []string
 	switch len(args) {
 	case 0:
 		_ = cmd.UsageFunc()(cmd)
@@ -348,9 +356,15 @@ func hostdbsetfiltermodecmd(cmd *cobra.Command, args []string) {
 		}
 	default:
 		filterModeStr = args[0]
-		for i := 1; i < len(args); i++ {
-			host.LoadString(args[i])
+
+		hostStrs := strings.Split(args[1], ",")
+		for _, hostStr := range hostStrs {
+			host.LoadString(hostStr)
 			hosts = append(hosts, host)
+		}
+
+		if len(args) > 2 {
+			netAddresses = strings.Split(args[2], ",")
 		}
 	}
 	err := fm.FromString(filterModeStr)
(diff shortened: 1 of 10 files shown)


node/api/hostdb.go
```diff
@@ -202,7 +205,7 @@ func (api *API) hostdbFilterModeHandlerPOST(w http.ResponseWriter, req *http.Req
 	}
 
 	// Set list mode
-	if err := api.renter.SetFilterMode(fm, params.Hosts); err != nil {
```
n8maninger, line 205: We should improve this API with add and delete endpoints instead of setting the list with each request.

modules/renter/hostdb/hostdb.go
```diff
@@ -39,6 +40,110 @@ var (
...
+func (bd *blockedDomains) managedRemoveDomains(domains []string) {
+	bd.mu.Lock()
+	defer bd.mu.Unlock()
+	for _, domain := range domains {
+		delete(bd.domains, domain)
+	}
+}
+
+// managedIsBlocked checks to see if the domain is blocked
+func (bd *blockedDomains) managedIsBlocked(addr modules.NetAddress) bool {
```
n8maninger, line 91: It'd be useful to also resolve the DNS host and check against its A and AAAA records.

modules/renter/hostdb/hostdb.go
```diff
@@ -39,6 +41,119 @@ var (
...
+}
+
+// managedAddDomains adds domains to the map of blocked domains
+func (bd *blockedDomains) managedAddDomains(domains []string) {
+	bd.mu.Lock()
+	defer bd.mu.Unlock()
+	for _, domain := range domains {
+		bd.domains[domain] = struct{}{}
+
+		addrs, err := bd.resolver.LookupHost(context.Background(), domain)
```
lukechampine, line 69: could we use `net.LookupHost` here instead? Seems like the benefit of using a distinct `net.Resolver` is that you can pass a context, but we don't make use of that here.

cmd/siac/hostdbcmd.go
```diff
@@ -37,11 +38,13 @@ var (
 	}
 
 	hostdbSetFiltermodeCmd = &cobra.Command{
-		Use:   "setfiltermode [filtermode] [host] [host] [host]...",
+		Use:   "setfiltermode [filtermode] [host,...] [netaddress,...]",
 		Short: "Set the filtermode.",
-		Long: `Set the hostdb filtermode and specify hosts.
+		Long: `Set the hostdb filtermode and specify comma separated hosts and net addresses.
```
lukechampine, line 43: perhaps instead, we could leave it as `[host] [host] [host]`, but allow `host` to be *either* a pubkey or a net address. That way, the existing behavior is preserved. It also avoids the ugliness of *only* specifying net addresses (which would otherwise require passing `""` for the list of pubkeys).

doc/api/index.html.md
```diff
@@ -2554,6 +2560,9 @@ Can be either whitelist, blacklist, or disable.
 **hosts** | array of strings  
 Comma separated pubkeys.  
 
+**netaddresses** | array of strings  
+Comma separated list of IP addresses or domains to block.
```
lukechampine, line 2564: strike the "to block", since this could be a whitelist

modules/renter/hostdb/hostdb.go
```diff
@@ -39,6 +41,119 @@ var (
 	errHostNotFoundInTree = errors.New("host not found in hosttree")
 )
 
+// blockedDomains manages a list of blocked domains
+type blockedDomains struct {
```
lukechampine, line 45: should be `filteredDomains`, since we support whitelisting as well

modules/renter/hostdb/hostdb.go
```diff
@@ -39,6 +40,110 @@ var (
...
+func (bd *filteredDomains) managedRemoveDomains(domains []string) {
+	bd.mu.Lock()
+	defer bd.mu.Unlock()
+	for _, domain := range domains {
+		delete(bd.domains, domain)
+	}
+}
+
+// managedIsFiltered checks to see if the domain is blocked
+func (bd *filteredDomains) managedIsFiltered(addr modules.NetAddress) bool {
```
n8maninger, line 99: Tests for these new functions would be good to add

chris124567:
Again everything works as expected.  Maybe the problem is the siatest test itself.

n8maninger:
> Again everything works as expected. Maybe the problem is the siatest test itself.

That's not too surprising; the `siatest` package has some very brittle tests. Unfortunately, I was also having issues figuring it out, so I added a new separate test for the net address filter and removed the subtests from `TestFilterMode`.  @chris124567 take a look and make sure it looks alright.

n8maninger:
> Again everything works as expected. Maybe the problem is the siatest test itself.

That's not too surprising; the `siatest` package has some very brittle tests. Unfortunately, I was also having issues figuring it out, so I added a new separate test for the net address filter and removed the subtests from `TestFilterMode`.  @chris124567 take a look and make sure it looks alright.

n8maninger:
I added a new test to `siatest/renter/hostdb` that creates a filter then announces a new host that should be filtered by the existing rule. It looks like we need to update the filtered list in `HostDB.insert` and `HostDB.modify` so that new and updated hosts are filtered immediately.

chris124567:
But don't modify and insert already check if a host is filtered?

```
    _, ok := hdb.filteredHosts[host.PublicKey.String()]
    isWhitelist := hdb.filterMode == modules.HostDBActiveWhitelist
```

chris124567:
I think this fixes it.  At least TestFilterMode, TestNetAddressFilter, TestNetAddressFilterContracts all pass on my machine.

chris124567:
This makes the behavior in modify match that in insert and SetFilterMode (only insert into filteredTree if whitelist==ok).

PR #107, opened 2022-04-28
miner module, siad and siac miner apis, siac seed command

alexfreska:


api/siad/api.go
```diff
@@ -37,3 +37,8 @@ type SyncerPeerResponse struct {
 type SyncerConnectRequest struct {
 	NetAddress string `json:"netAddress"`
 }
+
+// A MinerAddressRequest is for setting the miner's address.
+type MinerAddressRequest struct {
+	Address types.Address `json:"address"`
+}
```
lukechampine, line 44: not convinced this functionality is necessary. If you're a "real" miner, then yeah, you might want to change your address, but if you're running a CPU miner for testing, setting the address once at startup should be sufficient.

miner/miner.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// A Syncer can connect to other peers and synchronize the blockchain.
+	Syncer interface {
+		BroadcastBlock(block types.Block)
+	}
+)
+
+// MiningManager manages an internal block miner.
+type MiningManager struct {
```
lukechampine, line 31: I'd rather keep this code in `main` like it was before. Eventually we'll want to design a fancier mining package (for pools), but I imagine that will look very different from a package designed for CPU mining.

alexfreska:
> reworked this API a bit. Now, `siac` handles the mining instead of `siad`, and you submit mined blocks to `/consensus/broadcast`. This means we don't need stateful `/start` and `/stop` endpoints.
> 
> I also switched to a single `siac mine` command, which can be used to mine indefinitely (stopping when killed) or to mine a specific number of blocks. You can also specify the address that will receive the block reward.
> 
> Haven't tested this; can you try it out? (I'm starting to feel the need for an API test suite...)

@lukechampine I gave this a try and it works great, having it as a siac command with limit and address is very convenient. Yep a test suite would be great, if we get that set up I can help add good test cases as I consume new endpoints during UI development.

PR #108, opened 2022-04-29
Add file hosts function

n8maninger:
Adds a function to the renter that returns all hosts storing pieces of a file; request from Filebase.


lukechampine:
seems pretty inoffensive to me. What's up with our CI?

n8maninger:
> seems pretty inoffensive to me. What's up with our CI?

Looks like some sort of issue with the outdated version of Golangci-Lint we're using.

n8maninger:
Fixed by #110

PR #109, opened 2022-04-29
Fix sector removal lock contention

n8maninger:
Adds a sector removal map to the contract manager for removing large amounts of sectors. `RemoveSectorBatch` causes lock contention for long periods when removing large (2TB+) contracts and contracts with many virtual sectors due to individual sector locking/unlocking. When queueing sectors, only the map's mutex needs to be locked. When removed, physical and virtual sectors are batched into a single WAL operation. Each physical TiB takes around 3 hours to clear with the additional delay. Under ideal conditions, `RemoveSectorBatch` takes approximately 26 minutes per TiB due to a 500ms delay in the WAL sync; however, with lock contention, users have reported contracts with a few TiB taking days to remove.

The persistence file saves 20 bytes per physical sector, the 12-byte sector ID and an 8-byte counter. When a sector is queued, the counter is incremented. After removal, the counter is set to 0. All sectors with 0 counts are compacted when the host is reloaded to save storage space. 1TiB of physical sectors will be 5MiB on disk.

### DIFF
--- a/modules/host/contractmanager/consts.go
+++ b/modules/host/contractmanager/consts.go
@@ -49,6 +49,10 @@ const (
 	// sectorOverflowFile is the path to the file used if a virtual sector's
 	// counter becomes greater than the max value of a uint16.
 	sectorOverflowFile = "sector_overflow.dat"
+
+	// sectorRemovalFile is the path to the file used to store the sector removal
+	// queue.
+	sectorRemovalQueueFile = "sector_removal.dat"
 )
 
 const (
--- a/modules/host/contractmanager/sectorremoval.go
+++ b/modules/host/contractmanager/sectorremoval.go
@@ -27,9 +27,11 @@ type (
 	}
 
 	sectorRemovalMap struct {
-		mu          sync.Mutex
-		sectors     map[sectorID]removalEntry
-		entries     int64
+		mu sync.Mutex
+		// sectors tracks sector IDs that need to be removed
+		sectors map[sectorID]struct{}
+		// entries tracks the current state of the persistence file
+		entries     map[sectorID]removalEntry
 		removalFile *os.File
 
 		cm  *ContractManager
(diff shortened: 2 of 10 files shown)


modules/host/contractmanager/consts.go
```diff
@@ -49,6 +49,10 @@ const (
 	// sectorOverflowFile is the path to the file used if a virtual sector's
 	// counter becomes greater than the max value of a uint16.
 	sectorOverflowFile = "sector_overflow.dat"
+
+	// sectorRemovalFile is the path to the file used to store the sector removal
+	// queue.
+	sectorRemovalQueueFile = "sectorremoval.dat"
```
lukechampine, line 55: maybe `sector_removal.dat`, for parity with `sector_overflow.dat`?

modules/host/contractmanager/sectorremoval.go
```diff
@@ -0,0 +1,233 @@
...
+// and read errors are ignored. The file is compacted while loading.
+func (sq *sectorRemovalMap) managedLoad(path string) (err error) {
+	sq.mu.Lock()
+	defer sq.mu.Unlock()
+
+	sq.removalFile, err = os.OpenFile(path, os.O_RDWR|os.O_CREATE, modules.DefaultFilePerm)
+	if err != nil {
+		return errors.AddContext(err, "unable to open sector removal queue file")
+	}
+	if _, err := sq.removalFile.Seek(0, 0); err != nil {
```
lukechampine, line 58: shouldn't be necessary to seek immediately after opening

modules/host/contractmanager/sectorremoval.go
```diff
@@ -0,0 +1,233 @@
...
+		return errors.AddContext(err, "unable to seek to beginning of sector removal queue file")
+	}
+	buf := make([]byte, 20)
+	// read each entry from disk. The persistence file is compacted while
+	// reading.
+	for i := int64(0); ; i++ {
+		if _, err := sq.removalFile.ReadAt(buf, i*removalEntrySize); err == io.EOF || err == io.ErrUnexpectedEOF {
```
lukechampine, line 65: `ReadAt` implies we're reading out of order, but it looks like this is reading the whole file sequentially. Consider `io.ReadFull` instead.

A `bufio.Reader` seems warranted as well, since `removalEntrySize` is quite small.

modules/host/contractmanager/sectorupdate.go
```diff
@@ -375,6 +371,111 @@ func (wal *writeAheadLog) managedRemoveSector(id sectorID) error {
...
+	}
+
+	wal.appendChange(stateChange{
+		SectorUpdates: changes,
+	})
+
+	ch := wal.syncChan
+	wal.mu.Unlock()
+	wal.cm.sectorMu.Unlock()
+	<-ch
```
lukechampine, line 434: this certainly raised an eyebrow... why do we need to grab the WAL's channel here?

modules/host/contractmanager/sectorremoval.go
```diff
@@ -0,0 +1,233 @@
...
+		entries     int64
+		removalFile *os.File
+
+		cm  *ContractManager
+		wal *writeAheadLog
+	}
+)
+
+// writes the removal entry to disk
+func (sq *sectorRemovalMap) writeRemovalEntry(id sectorID, count uint64, offset int64) error {
```
lukechampine, line 40: prefer to match receiver names to type names (`srm`)

modules/host/contractmanager/sectorupdate.go
n8maninger: It's a holdover from `managedRemoveSector` to wait until after the WAL has synchronized to update the sector metadata and clear the usage.

modules/host/contractmanager/sectorremoval.go
```diff
@@ -0,0 +1,232 @@
...
+		// skip entries that are no longer needed.
+		if entry.Count == 0 {
+			continue
+		}
+
+		srm.sectors[entry.ID] = entry
+		// if the number of read entries is greater than the number of used
+		// entries overwrite the unused entries.
+		if srm.entries != i {
+			if err := srm.writeRemovalEntry(entry.ID, entry.Count, entry.Offset); err != nil {
```
lukechampine, line 84: is the idea here that all of the entries "shift down" to fill in the gap? That would be O(n) in the number of entries, right? Should we be concerned about that?

modules/host/contractmanager/sectorremoval.go
```diff
@@ -0,0 +1,232 @@
...
+				return errors.AddContext(err, "error updating sector removal entry")
+			}
+		}
+		// increment the number of entries
+		srm.entries++
```
lukechampine, line 89: does `srm.entries` always equal `len(srm.sectors)`?

modules/host/contractmanager/sectorremoval.go
n8maninger: Yes, all entries will shift down to fill gaps; it makes tracking the file offset of entries relatively simple. I don't think we need to be concerned since it only happens when the file is loaded. We could batch the writes to reduce syscalls or add a temp file+rename at the end of the load.

n8maninger:
Disabled the linter error for `defer File.Close()`

PR #112, opened 2022-05-04
Add send siacoin to siac wallet

alexfreska:


api/siad/server.go
```diff
@@ -69,6 +70,9 @@ func (s *server) consensusStateHandler(w http.ResponseWriter, req *http.Request,
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
+	// TODO: add marshalling to accumulators
+	cs.Elements = merkle.ElementAccumulator{}
+	cs.History = merkle.HistoryAccumulator{}
```
alexfreska, line 75: These are missing, I was going to add an API response struct that omits these, but we are calling `InputSigHash` on the `siac` side.

lukechampine: not sure I understand -- what exactly is missing? Both of these types have `MarshalJSON`/`UnmarshalJSON` methods.

lukechampine:
@alexfreska I refactored this a bit, can you test it again and make sure I didn't break anything?

alexfreska:
> @alexfreska I refactored this a bit, can you test it again and make sure I didn't break anything?

@lukechampine I just tried out the latest changes and was able to send sc transactions, looks good 👍🏼

PR #116, opened 2022-05-17
v2: add a basic HostDB

chris124567:
Adds a basic HostDB with an ephemeral store and a store that supports writing JSON to disk.


renter/hostdb/hostdb.go
```diff
@@ -0,0 +1,71 @@
...
+func (db *DB) ProcessChainRevertUpdate(cru *chain.RevertUpdate) error {
+	for _, txn := range cru.Block.Transactions {
+		for _, a := range txn.Attestations {
+			if a.Key == announcementAttestationKey {
+				if err := db.store.RemoveHost(a.PublicKey); err != nil {
+					return err
+				}
+			}
+		}
+	}
```
lukechampine, line 59: you can factor this out into a `func findAnnouncements(b types.Block) []Host`

PR #119, opened 2022-05-26
v2: Add HostDB scanning and weighting

chris124567:
This PR ports much of the v1 hostdb code to v2.  Not all of the weighting adjustments were ported due to some of them relying on things that don't exist in v2 (like a base RPC price).  The hostdb also still needs to be integrated into the renter.


renter/hostdb/weight.go
```diff
@@ -0,0 +1,465 @@
...
+		return 0
+	}
+	return db.acceptContractAdjustments(host) *
+		db.lifetimeAdjustments(host) *
+		// db.basePriceAdjustments(host) *
+		db.collateralAdjustments(host, allowance) *
+		db.durationAdjustments(host, allowance) *
+		db.interactionAdjustments(host) *
+		// db.priceAdjustments(host, allowance) *
+		// db.storageRemainingAdjustments(host, allowance) *
```
lukechampine, line 462: I think we can get away with a small set of requirements instead of a full-blown `Allowance` type.
Maybe an API like:
```go
func Random(n int, reject func(Host) bool) []Host
```
`reject` can be a fairly simple function that checks whether the `WindowSize` is too small and whether the collateral/price is acceptable. That way, the hostdb doesn't need to have any concept of an allowance.

renter/hostdb/weight.go
```diff
@@ -0,0 +1,465 @@
...
+	return 1
+}
+
+// lifetimeAdjustments will adjust the weight of the host according to the total
+// amount of time that has passed since the host's original announcement.
+func (db *DB) lifetimeAdjustments(host Host) float64 {
+	index, err := db.store.Index()
+	if err != nil {
+		return math.SmallestNonzeroFloat64
+	}
```
lukechampine, line 60: better to retrieve this value in `Score`, and pass it to `lifetimeAdjustments` as a parameter.

As a general note, I suggest making these functions standalone, rather than methods on `db`.

renter/hostdb/scan.go
```diff
@@ -0,0 +1,288 @@
...
+		sleepRange := uint64(maxScanSleep - minScanSleep)
+		sleepTime := minScanSleep + time.Duration(frand.Uint64n(sleepRange))
+
+		// Sleep until it's time for the next scan cycle.
+		select {
+		case <-db.tg.StopChan():
+			return
+		case <-time.After(sleepTime):
+		}
+	}
```
lukechampine, line 287: so complicated! Let's do something like this (pseudocode) instead:
```go

func (db *DB) scanLoop() {
	scanCh := make(chan struct{}, maxInflightScans)
	for i := 0; i < maxInflightScans; i++ {
		go func() {
			for {
				host := <-scanCh
				res := scanHost(host)
				db.ModifyHost(host, res)
			}
		}()
	}

	for {
		host, err := store.HostForScan(since)
		if err == NoUnscannedHosts {
			// sleep for a while
			time.Sleep(aWhile)
			continue
		}
		scanCh <- host
	}
}
```

renter/hostdb/update.go
```diff
@@ -0,0 +1,107 @@
+package hostdb
+
+import (
+	"math"
+
+	"go.sia.tech/core/types"
+)
+
+func updateHostHistoricInteractions(host *Host, index types.ChainIndex) {
```
lukechampine, line 9: I question whether we should use a `ChainIndex` here. It seems like what we really care about is *time*, and we're just using the block height as a stand-in for that.

chris124567:
Implemented all of the suggestions except I was a little confused how we want to handle the allowance/rejection stuff.  For instance collateralAdjustments uses information from the allowance not to reject hosts (i.e. return math.SmallestNonzeroFloat64) but to provide a scoring adjustment based off the collateral the host provides.  Here is what I have so far: https://github.com/SiaFoundation/siad/pull/119/commits/8a3c01dfb00a88001021f5de0dbe2ef97d3b4b21#diff-75cdc0be7c88ff2822ed700103d88b9fc0748d5589b41e6979358d5eca0fbb7dR393

PR #123, opened 2022-06-11
Remove need for /api prefix in api routes

chris124567:
This PR allows us to declare the API routes without the /api prefix but serve them under that route in the various binaries (siad, renterd, explored).

### DIFF
--- a/cmd/explored/main.go
+++ b/cmd/explored/main.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"time"
 
 	"go.sia.tech/core/consensus"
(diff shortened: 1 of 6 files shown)


cmd/explored/main.go
```diff
@@ -85,7 +86,14 @@ func main() {
...
 	go func() {
 		api := explored.NewServer(n.c, n.s, n.tp)
-		if err := http.Serve(l, api); err != nil {
+		if err := http.Serve(l, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if strings.HasPrefix(r.URL.Path, "/api") {
+				r.URL.Path = strings.TrimPrefix(r.URL.Path, "/api")
+				api.ServeHTTP(w, r)
+				return
+			}
+			http.NotFound(w, r)
```
lukechampine, line 95: does this work?

```go
mux := http.NewServeMux()
mux.Handle("/api/", explored.NewServer(n.c, n.s, n.tp))
if err := http.Serve(l, mux); err != nil {
```

chris124567: http://127.0.0.1:9980/api/txpool/transactions 404s with this

PR #124, opened 2022-06-13
Extend subscription rhp with ability to subscribe by registry entry id

ChrisSchinnerl:
This MR adds another type of request to the subscription rhp to allow for subscribing only via registry ID.
The result is the same as for regular subscriptions except for the initial values we send to the client after subscribing. These now contain the public key to allow the client to verify that the responses belong to the values they requested.

### DIFF
--- a/modules/host/rpcsubscribe.go
+++ b/modules/host/rpcsubscribe.go
@@ -138,9 +138,53 @@ func (h *Host) managedHandleSubscribeRequest(info *subscriptionInfo, pt *modules
 		}
 		rvs = append(rvs, rv)
 	}
+	if err := h.managedHandleFinalizeSubscribeRequest(info, pt, ids, uint64(len(rvs))); err != nil {
+		return err
+	}
 
+	// Write initial values to the stream.
+	return errors.AddContext(modules.RPCWrite(stream, rvs), "failed to write initial values to stream")
+}
+
+// managedHandleSubscribeByRIDRequest handles a new subscription.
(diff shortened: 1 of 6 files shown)


modules/host/rpcsubscribe.go
```diff
@@ -193,6 +232,39 @@ func (h *Host) managedHandleUnsubscribeRequest(info *subscriptionInfo, pt *modul
 	return errors.AddContext(err, "failed to signal successfully unsubscribing from entries")
```
n8maninger, line 232: This should also use `managedHandleFinalizeUnsubscribeRequest`

n8maninger (commented):
Two small things we could address in a follow-up. @lukechampine, should we merge this and get an RC1 out this afternoon?

modules/host/rpcsubscribe.go
ChrisSchinnerl: done

modules/renter/workersubscription_test.go
```diff
@@ -1298,3 +1298,122 @@ func TestThreadedSubscriptionLoop(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestSubscribeUnsubsribeByRID tests the subscription helper methods against
+// the worker tester. They are already unit-tested against a host in
+// rpcsubscribe_test.go but better safe than sorry.
+func TestSubscribeUnsubsribeByRID(t *testing.T) {
```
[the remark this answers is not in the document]
ChrisSchinnerl: and done

lukechampine:
TestNewProgramLowCollateralBudget is failing -- any connection to this?

n8maninger:
> 

Looks like another NDF

Issue #127, opened 2022-06-18
siac update not working on Windows


Cerlancism:
Related discussion on Discord: https://discord.com/channels/809849352516141067/827641929008152635/987703237513719838

![image](https://user-images.githubusercontent.com/28769707/174451297-8ba0ad89-577e-4188-a83a-8680b9f5af3a.png)

It seems this line does not account for `.exe` as binary executable file ending on Windows, and also missing the `return` for error case "No checksum found for binary".
https://github.com/SiaFoundation/siad/blob/821837bc1a1830a654f39c5884ee30a0c51df694/node/api/daemon.go#L279-L285

PR #128, opened 2022-06-19
Use binaryName for siac update checksum and return error for no checksum found

Cerlancism:
My attempt to fix https://github.com/SiaFoundation/siad/issues/127


lukechampine:
Need to run gofmt, but otherwise this LGTM

lukechampine:
thanks!

PR #129, opened 2022-06-20
Remove "/api" prefix from client functions

chris124567:
Makes Client.req add the "/api" prefix after BaseURL.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -22,7 +22,7 @@ func (c *Client) req(method string, route string, data, resp interface{}) error 
 		js, _ := json.Marshal(data)
 		body = bytes.NewReader(js)
 	}
-	req, err := http.NewRequest(method, fmt.Sprintf("%v%v", c.BaseURL, route), body)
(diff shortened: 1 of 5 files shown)


api/api.go
```diff
@@ -22,7 +22,7 @@ func (c *Client) req(method string, route string, data, resp interface{}) error
 		js, _ := json.Marshal(data)
 		body = bytes.NewReader(js)
 	}
-	req, err := http.NewRequest(method, fmt.Sprintf("%v%v", c.BaseURL, route), body)
+	req, err := http.NewRequest(method, fmt.Sprintf("%v/api%v", c.BaseURL, route), body)
```
lukechampine, line 25: nah, not even here. Consider that nothing prevents someone from serving the API under a different prefix; `Client` would then be unable to access it. Also consider that, if you're making requests using `curl`, you'll manually specify the `/api` prefix; it would be confusing if that were not also the case for `Client`. It should be the caller's responsibility to pass the correct `BaseURL`.

chris124567:
:+1:

PR #137, opened 2022-07-05
Update blockedWithdrawalTimeout

ChrisSchinnerl:


modules/host/accountmanager.go
```diff
@@ -63,8 +63,8 @@ var (
 	// blockedWithdrawalTimeout is the amount of time after which a blocked
 	// withdrawal times out.
 	blockedWithdrawalTimeout = build.Select(build.Var{
-		Standard: 15 * time.Minute,
-		Dev:      5 * time.Minute,
+		Standard: 3 * time.Minute,
```
peterjan, line 66: I would drop a `NOTE` in the docstring here indicating how it's related to the siamux default timeouts

ChrisSchinnerl: added

PR #141, opened 2022-07-07
Use the host tg's stop context instead of using the OnStop callback.

Didrole:
This leaked the created context until host shutdown.


lukechampine:
Thanks!

PR #144, opened 2022-07-19
Unlock wallet before other modules startup

n8maninger:
Moves the wallet auto unlock before other modules have loaded to prevent host proof issues
```
storageobligations.go:1351: failed to build storage proof trransaction: Host error when funding a storage proof transaction fee: wallet must be unlocked before it can be used
```

### DIFF
--- a/cmd/siad/daemon.go
+++ b/cmd/siad/daemon.go
@@ -182,19 +182,6 @@ func installKillSignalHandler() chan os.Signal {
 	return sigChan
 }
 
-// tryAutoUnlock will try to automatically unlock the server's wallet if the
-// environment variable is set.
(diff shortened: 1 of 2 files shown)


n8maninger:
Added ignores for the new gosec errors since they're not from new code:
```
[lint: types/currency.go#L243] G113: Potential uncontrolled memory consumption in Rat.SetString (CVE-2022-23772) (gosec)
```
```
[lint: node/api/server_helpers_test.go#L135] G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
```
```
[lint: cmd/siac/parse.go#L97] G113: Potential uncontrolled memory consumption in Rat.SetString (CVE-2022-23772) (gosec)
```

PR #151, opened 2022-11-13
Docker timezone support fix

timrettop:
To support setting the timezone for siad docker image the docker build needs tzinfo package, a TZ environment variable set, and the zoneinfo data from the build copied. 

This increases the image size from 33.3MB to 34.4MB

Tested by building and running an image with submitted Dockerfile using TZ env variable and checking `date` output. 
FYI, also attempted to set volume mounts on existing published image for /etc/timezone /etc/localtime and /usr/share/zoneinfo and could not get the image to detect the TZ env variable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY . .
 
 # need to run git status first to fix GIT_DIRTY detection in makefile
 RUN apk update \
-	&& apk add --no-cache build-base git make ca-certificates \
+	&& apk add --no-cache build-base git make ca-certificates tzdata \
 	&& update-ca-certificates \
 	&& git status > /dev/null \
 	&& make static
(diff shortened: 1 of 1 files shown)


kalvdans:
Why does siad need timezones?

timrettop:
> Why does siad need timezones?

For easier log reading and searching

n8maninger:
This doesn't appear to affect `siad`'s logs:
```
/sia-data/gateway # echo $TZ
America/Chicago
/sia-data/gateway # date
Sun Nov 13 10:35:36 CST 2022
/sia-data/gateway # cat gateway.log | head -n 1
2022/11/13 16:28:02.148713 log.go:38: STARTUP: Logging has started. siad Version 1.5.9
```

The external logging package, `gitlab.com/NebulousLabs/log`, sets the `LUTC` flag on the logger; all log lines will use UTC.

timrettop:
Darn, I should have looked further in testing.  Ahh well

PR #153, opened 2023-01-12
Add Testnet

n8maninger:
+ Adds a `Testnet` field to `build.Var`
+ Adds "testnet" to `build.Release` 
+ Adds testnet const overrides, most of them should mirror the standard values
+ Moves all the hard forks within the first 20 blocks,  maintaining the original ordering
+ Adds a `siad-testnet` docker image
+ Adds a large SCT? airdrop to genesis for a faucet
+ Sends the initial SFT? distribution to the Foundation wallet
+ Disables the fork at block `10000` moving SFT? to the dev address for testnet builds


lukechampine:
All those `build.Var`s 😭 
Too late now, but an alternative would be to tweak `build.Select` to use the `Standard` value for `Testnet` if `Testnet` is not defined. Then you'd only need to add `Testnet` values if they were different.

It would be nice if we could ensure that testnet addresses were different from mainnet addresses (Bitcoin does this), but our address format doesn't include a version byte so I don't think it's possible. Bitcoin's testnet also resets the difficulty if no blocks have been found in a while, but I'm not sure how it's implemented.

n8maninger:
> an alternative would be to tweak build.Select to use the Standard value for Testnet

I initially did it this way but ended up changing it. It made it easier to miss the few vars that did need to be changed.

PR #158, opened 2023-03-29
Put a cap on the expectedExchange within verifyClearingRevision

ChrisSchinnerl:
To avoid contracts failing to renew when they are completely drained, I put a cap on the `expectedExchange` in `verifyClearingRevision`.

rhpv3 doesn't suffer from that issue since we always pass 0 but in rhpv2 we might fail to renew drained contracts which is not something that either renter nor host would want on a well-used contract.


n8maninger:
Let's leave the existing logic. That will mean heavily used `siad` contracts will continue to fail to renew, but we're not planning on another release regardless.

ChrisSchinnerl:
Closing this in favour of upgrading to rhpv3

PR #165, opened 2023-08-01
Fix write to a nil map during contract maintenance

mike76-dev:
I know that `siad` will only be receiving critical updates. However, even if this has been working fine so far, we might be catching edge cases, or any forks of `siad` might be catching this.


n8maninger:
From a quick glance, it looks like `numFailedRenews` is assigned in `threadedContractMaintenance` before any write calls. Did you actually hit a panic? I'm not opposed to merging. My only gripe would be putting back all the `types.BlockHeight`. It's annoying, but better to be consistent when there's no real reason to change it.

mike76-dev:
I hit a deadlock when running contract maintenance. While debugging, I found that the map could be read from but couldn't be written to.
Regarding the types, I will put them back; however, I find it weird to use `types.BlockHeight` when semantically it's just a counter that is incremented every time a renewal fails.

